PR #43, opened 2015-01-08
Pull a110x

Jon-ICS:
This adds support for the a110x Hall Effect sensor.


sknepper:
Beautiful!  Merged in, so I'll close the PR.  Thanks as always!

PR #49, opened 2015-01-09
groveelectricity: Initial implementation

Jon-ICS:
This module adds support for the Grove Electricity sensor.


sknepper:
Does it make sense to rename the library to the more general TA12200, as the Grove Electricity sensor is based on the TA12-200 current transformer?  If so, I can make the changes when merging in.

Jon-ICS:
On 01/15/2015 11:51 AM, sknepper wrote:
>
> Does it make sense to rename the library to the more general TA12200, 
> as the Grove Electricity sensor is based on the TA12-200 current 
> transformer? If so, I can make the changes when merging in.
>

I'll leave that up to you - I wasn't sure how ubiquitous that device 
was, so I elected to be specific to grove.

-- 
Jon Trulson
Integrated Computer Solutions
www.ics.com

src/groveelectricity/groveelectricity.cxx
```diff
@@ -0,0 +1,113 @@
...
+#include <iostream>
+
+#include "groveelectricity.h"
+
+using namespace upm;
+using namespace std;
+
+GroveElectricity::GroveElectricity(int pin)
+{
+  mraa_init();
```
arfoll: Great practice to call mraa_init() but if you don't check the return value then it was useless.

Jon-ICS:
This is true, unless it succeeded :)  

Sarah, do you want me to do a rename to TA12200 and add a check for the return value of mraa_init() as well?  I notice in some of my earlier drivers, I neglected to do this as well (like the grove reed relay).  I can look over the outstanding PR's and check to make sure mraa_init() is being checked properly.

arfoll:
mraa_init() has already been run because of the gcc ctor compiler hint in mraa (and your libc has to support ctors since this is C++ code). So the only reason to re-run it in this case is to check that it didn't fail :). See the mraa internal docs for more details.

Jon-ICS:
On 01/15/2015 01:11 PM, Brendan Le Foll wrote:
>
> mraa_init() has already been run because of the gcc ctor compiler hint 
> in mraa (and your libc has to support ctors since this is C++ code). 
> So the only reason to re-run it in this case is to check that it 
> didn't fail :). See the mraa internal docs for more details.
>

I think I was basing that (calling mraa_init()) off of the original 
grove sensors (src/grove) which also calls mraa_init() without 
checking.  I can just remove this call if it's not neccessary.

-- 
Jon Trulson
Integrated Computer Solutions
www.ics.com

sknepper:
I can do the rename and remove the mraa_init() call, but thanks for the offer!

sknepper:
Merged in - thanks much!

PR #56, opened 2015-01-09
ublox6: Initial implementation

Jon-ICS:
The module implements support for the ublox-6 GPS sensor.  It was
tested on a Grove GPS device.

This module simply allows data to be retrieved from the device in the
form of NMEA sentences, and provides a method to write commands to the
device for configuration purposes.

It does not attempt to parse NMEA data -- that is a project in itself.
There are libraries available on the Internet, such as tinyGPS++ that
can handle that for you.


jrvandr:
Thanks!  Merged and closing.

PR #59, opened 2015-01-12
mq303a: Initial implementation of the mq303a alcohol sensor

Jon-ICS:
The module implements support for the mq303a alcohol sensor.  It was
tested on the Grove Alcohol Sensor.


src/mq303a/mq303a.h
```diff
@@ -0,0 +1,76 @@
...
+    MQ303A(int pin, int heaterPin);
+
+    /**
+     * MQ303A Destructor
+     */
+    ~MQ303A();
+
+    /**
+     * Get the alcohol detected from the sensor.  This value is inverted,
+     * a lower value means more alcohol detected.
```
sknepper: This comment seems a little misleading.  In the library, the inversion is done (1023 - mraa_aio_read(m_aio)) for the user.  In the examples, it says that a higher value from calling mq303a->value() means stronger alcohol.
Would a more appropriate comment for this function be something like:
"The value read from the analog pin is inverted.  A higher returned value means a higher amount of alcohol was detected."

Jon-ICS: Yes... This comment is confusing.  Yours is better.  In addition, the @return tag is also not correct for this sensor.  I can fix both if you'd like.

sknepper: Thanks for the confirmation!  I can make the changes when merging (I noticed the @return tag for this function as well as the next one, and made the changes locally already).  Thanks for the offer, though.  =)

Jon-ICS: No problem... Careful with the @ mentions though - someone named joseph hill has that name on github and is probably getting these messages :)

sknepper:
Merged on in with the changes described.  Thanks much!

PR #67, opened 2015-01-12
lcd: JavaScript example for RGB LCD in Grove starter kit

Jon-ICS:

### DIFF
--- a/examples/javascript/rgblcd.js
+++ b/examples/javascript/rgblcd.js
@@ -1,0 +1,51 @@
+/*jslint node:true, vars:true, bitwise:true, unparam:true */
+/*jshint unused:true */
+/*global */
+/*
+ * Author: Zion Orent <zorent@ics.com>
+ * Copyright (c) 2014 Intel Corporation.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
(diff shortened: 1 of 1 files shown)


examples/javascript/rgblcd.js
```diff
@@ -0,0 +1,51 @@
...
+ * WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+var mraa = require('mraa'); //require mraa
+
+// Load lcd module on I2C
+var LCD = require('jsupm_i2clcd');
+// Initialize Jhd1313m1 at 0x62 (RGB_ADDRESS) and 0x3E (LCD_ADDRESS) 
+
+// Look at this link to see where the BUS ID fix for Edison came from:
+//	https://software.intel.com/en-us/forums/topic/533842
```
sknepper: At one point, this workaround was indeed needed.  However, as of at least Nov. 5 (using a posting date from the forum link given), it was fixed in a version of MRAA.  To avoid confusion, I'd prefer using bus 0 everywhere since it seems to work - I just tested on my Edison board using bus 0 and it was fine.  Do you mind if I make that change when merging this in?

Jon-ICS:
Not at all.  We may have been using on older version of mraa at the time.

arfoll:
@Jon-ICS getPlatformName() is much more recent than the i2c bus enumeration change so your version of mraa can't have been that old. Just use LCD.Jhd1313m1(0) that will always work or at least *a* i2c bus will do something if one is present because of the way we enumerate them in mraa's platform config.

sknepper:
Merged in!  I  changed the name to rgb-lcd.js (with a hyphen) to be consistent with the name of the existing C++ example.  Thanks!

PR #69, opened 2015-01-12
ssd1327: JavaScript example for ssd1327 OLED display

Jon-ICS:


sknepper:
Split into a few commits (since there's a new JavaScript example, fixes to the C++ example, and the swig changes as well).  Merged in, thanks so much!

PR #75, opened 2015-01-15
cmake: allows to build in source tree by filtering subdirs

rzr:
Without it FTBFS and displays this error message :

  cmake . && make
  (...)
  The source directory
  (...) src/CMakeFiles
  does not contain a CMakeLists.txt file.

Change-Id: I08efc4667d1004a5d19575dd4464dcd89d03fb28
Signed-off-by: Philippe Coval <philippe.coval@open.eurogiciel.org>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,10 @@
 cmake_minimum_required (VERSION 2.8.11)
 project (upm)
 
+if ("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
+message("WARNING: building into sources dir can be risky, prefer other directory")
+endif ()
+
 find_package (SWIG)
 if (SWIG_FOUND)
   include (${SWIG_USE_FILE})
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -125,5 +125,7 @@ endmacro(upm_module_init)
 
 subdirlist(SUBDIRS ${CMAKE_CURRENT_SOURCE_DIR})
 foreach(subdir ${SUBDIRS})
-    add_subdirectory(${subdir})
+    if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/${subdir}/CMakeLists.txt)
+        add_subdirectory(${subdir})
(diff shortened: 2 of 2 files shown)


arfoll:
@sknepper @Propanu Can you at least advise when this fix is likely to go in?

Propanu:
Honestly, this was the first thing I tried when I started with MRAA/UPM. However, I quickly realized that doing an out-of-source build is the way to go. If you build in the source tree, you'll fill the sources with objects and create a mess. Also, for me it broke down after a few builds due to recursive dir generation and had to start over from a fresh clone. I had different changes but they achieved the same thing as your "fix":
```
foreach(child ${children})
    if(IS_DIRECTORY ${curdir}/${child} AND NOT ${child} STREQUAL "CMakeFiles")
        set(dirlist ${dirlist} ${child})
    endif()
```
Thus I would leave the CMakeLists as they are, and look at the errors as a way to enforce out-of-source builds, saving you some trouble in the long run. @sknepper @jrvandr

rzr:
Well some build system are building into sources tree  ... 

Anyway it was merged downstream in tizen and working fine since there is no persistance of generated files that create the mess ...

May I just add a warning message that suggest user to build out of sources but allow this ?

It's prefered to avoid downstream changes since I tend to forward upstream first.

Thanks for feedback anyway

arfoll:
@Propanu @rzr out of tree builds are definately not recommended however some build systems like to do it (they chuck the src tree every time so mess/broken after first build isn't an issue). Does obs always do out of tree builds for cmake projects? Yocto does it outside of the git clone which is annoying because our cmake git tag module doesn't work :/

IMHO the fix doesn't harm anything, it provides the extra safety of never adding a folder that doesn't have a cmakelists file so I don't see any reason not to include.

Propanu:
Please add this and I'll have no problem merging the changes.
> May I just add a warning message that suggest user to build out of sources but allow this ?

Propanu:
Thanks, merged as https://github.com/intel-iot-devkit/upm/commit/a665b5d6f1db35531ac7d6b2ce7d8bd55f04c596

PR #76, opened 2015-01-16
mq2: JavaScript example, with appropriate SWIG typemap

Jon-ICS:


sknepper:
Wonderful!  Split into two commits because the SWIG typemap is common to all the gas files (not just for mq2).  Thanks much!

PR #81, opened 2015-01-17
Ppd42ns fixes - patch previously broke compilation on systems with glibc < 2.17

arfoll:
See man 2 clock_gettime for more details.


sknepper:
Thanks so much for the fix(es)!  The messages are clear to understand the changes needed.  Thank you!!

PR #85, opened 2015-01-19
itg3200: Modifications to swig JavaScript and a new JavaScript example

sknepper:


sknepper:
I merged in, so I'll close.

PR #88, opened 2015-01-21
Itg3200 py

jrvandr:
Python sample along with swig python pointer changes


Propanu:
Can you please do a quick rebase and squash these into a single commit?

sknepper:
Actually, I think it's better to keep them separated.

sknepper:
One of the commits has to do with fixing the Python bindings, the other is a new example.  For previous PRs like this (with a new example and also a fix for the bindings), we've been separating them into multiple commits.  Only when the entire sensor library is new have we been squashing commits into just one commit.

jrvandr:
merged and closing

PR #91, opened 2015-01-22
Pull wt5001 use mraa_uart_get_dev_path

Jon-ICS:
The request updates the driver to use mraa_uart_get_dev_path().

It also renames the JS example from w5001 -> wt5001, like the C++ example.   

Then the JS example is fixed to support the new driver interface.


Propanu:
https://github.com/intel-iot-devkit/upm/commit/06b48d8e0783faa7a5a516d192f6323234d9fc56

PR #103, opened 2015-02-04
lcd jhd1313m1 clear() patch - added usleep after i2Cmd

jrvandr:
This patch fixes issue with putting a clear() right before a write()
resulting in loss of characters on the write.  The usleep delays for
a short time then the write() data is displayed properly.

Signed-off-by: John Van Drasek <john.r.van.drasek@intel.com>


Propanu:
Issue is with write function, sleep should be inserted there instead!

Issue #110, opened 2015-02-13
upm::Servo::setAngle has incorrect logic for servo positioning


tophyr:
The `setAngle` method in `upm::Servo` uses incorrect logic to position the servo it controls. It appears to send a PWM signal for a calculated number of 8ms periods, then shut off the PWM signal entirely. This behavior is incorrect in that the servo does not have time to move to its intended location, and even if it did the servo would not hold its location once it arrived.

jrvandr:
Thank you for contributing this bug fix!  You're pull request has been merged in.

PR #111, opened 2015-02-13
Fix Servo library so that it correctly moves to the given angle.

tophyr:
Fixes issue #110


jrvandr:
Many thanks for the contribution!  Merged and closing.

PR #117, opened 2015-02-25
Add createChar method to I2CLcd class

0xD34D:
LCDs based on the HD44780 allow for setting 8 custom characters.
This patch adds a method that exposes this capability.

### DIFF
--- a/src/lcd/i2clcd.cxx
+++ b/src/lcd/i2clcd.cxx
@@ -48,6 +48,20 @@ I2CLcd::write (int row, int column, std::string msg) {
 }
 
 mraa_result_t
+I2CLcd::createChar(uint8_t charSlot, uint8_t charData[]) {
+    mraa_result_t error = MRAA_SUCCESS;
+    charSlot &= 0x07; // only have 8 positions we can set
+    error = i2Cmd(m_i2c_lcd_control, LCD_SETCGRAMADDR | (charSlot << 3));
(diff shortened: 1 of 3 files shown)


Propanu:
Thank you for your contribution!

We do have a derived class for HD44780 based displays (lcm1602). I know the name is not very intuitive but could you please place your new function in there instead?

Also, to make things faster please consider using our conventions from contributions.md:
- signing-off your commits
- name them "<module>: description"

0xD34D:
So that would mean this code would also need to be duplicated in Jhd1313m1 and any other displays in the lcd module as well.  Jhd1313m1 inherits from I2CLcd and would not be able to benefit from this change.

It made more sense to include this in a base class that can be used in other classes such as lcm1602 and Jhd1313m1.  FYI, Jhd1313m1 was what I have been testing with since I have the Grove LCD with RGB backlight.

Propanu:
Sorry, wasn't aware the Grove LCD (jhd1313m1) is also HD44780 based. It's fine to add your changes in the base class for now, but please add a sign-off line to your commits!
Thanks!

0xD34D:
Commits updated with module name and signed-off-by.

Propanu:
Thanks a lot!
Merged in f50ab24bb3eacff79ba0f48f2fb9f67b77afc15c

PR #128, opened 2015-03-09
groveeldriver: initial implementation

Jon-ICS:
This module implements support for the Grove EL Driver.


examples/javascript/groveeldriver.js
```diff
@@ -0,0 +1,54 @@
...
+// When exiting: turn off EL wire, clear interval, and print message
+process.on('SIGINT', function()
+{
+	clearInterval(myInterval);
+	eldriver_obj.off();
+	eldriver_obj = null;
+	eldriver_lib.cleanUp();
+	eldriver_lib = null;
+	console.log("Exiting...");
+	process.exit(0);
```
rwaldron: None of this explicit memory clearing code is necessary when running JavaScript on node.js: V8 will free any allocated resources when the program is terminated. `null`'ing hasn't been necessary in any JavaScript runtime since Internet Explorer 8 (required `null`'ing dom elements that had been removed from the tree, regardless of reference reachability). Invoking V8's GC is pointless here because V8 is going to clean up all of the memory _anyway_.

Generally, there is no need to expose V8's GC via this (or any) library—V8 will run GC when it needs to. Unless the program is collecting heap data and needs to intentionally trigger GC, then user code never needs access to V8's GC.

Zion-ICS: Unfortunately, v8 doesn't guarantee garbage collection.
https://code.google.com/p/v8-juice/wiki/CommentaryOnV8#No_GC_guaranty

In testing upm, we found that GC wasn't called-- neither on scope exit nor on program exit. This is why we call it manually in the examples.

I wouldn't make a 1-to-1 comparison between JavaScript behavior in a browser and JavaScript behavior in nodeJS. Pretty much all browsers perform correct GC, and browser JavaScript likely hasn't required any type of variable destruction for quite a while.

But nodeJS is the wild west regarding GC. As of node 0.10.0, any attempt at guessing when to collect garbage was removed. This was because nodeJS couldn't figure out how to call GC in an efficient manner (see "Latency and Idle Garbage Collection" in the link).
http://blog.nodejs.org/2013/03/11/node-v0-10-0-stable/
This leaves us with V8 GC-- which isn't guaranteed.

Essentially, v8 was built by Google for Google Chrome. Everybody else using v8 has to deal with issues like this one from time to time.

rwaldron: > In testing upm, we found that GC wasn't called-- neither on scope exit nor on program exit. 

This would be an egregious memory leak, but not one that's the result of running a JavaScript program via node.js. Consider tools like Forever (or anything similar) that automatically restart node.js programs that have terminated unexpectedly: if memory isn't implicitly freed on termination, repeated restarts would cripple the system.


> I wouldn't make a 1-to-1 comparison between JavaScript behavior in a browser and JavaScript behavior in nodeJS. 

Yes, neither would I. I've also been actively developing software for node.js since 2010—so I have a significant amount of experience with the platform. 


> Pretty much all browsers perform correct GC, and browser JavaScript likely hasn't required any type of variable destruction for quite a while.

The "browser" has nothing to do with the JavaScript runtime's GC behaviour, aside from providing the runtime (V8, SpiderMonkey, JCS, Chakra, etc) with metrics that inform its scheduling and execution of a GC cycle. I gave the example of IE8 to illustrate the pointlessness of exposing and invoking GC explicitly. 


> But nodeJS is the wild west regarding GC. As of node 0.10.0, any attempt at guessing when to collect garbage was removed. 

From the [release announcement, almost exactly two years ago](http://blog.nodejs.org/2013/03/11/node-v0-10-0-stable/) that you cited: 

>> One of the toughest things to get right in a garbage collected language is garbage collection. In order to try to avoid excessive memory usage, Node used to try to tell V8 to collect some garbage whenever the event loop was idle.

>> However, knowing exactly when to do this is extremely difficult. There are different degrees of "idleness", and if you get it wrong, you can easily end up spending massive amounts of time collecting garbage when you'd least expect. In practice, disabling the IdleNotification call yields better performance without any excessive memory usage, because V8 is pretty good at knowing when it's the best time to run GC.

>> So, in v0.10, we just ripped that feature out. (According to another point of view, we fixed the bug that it was ever there in the first place.) As a result, latency is much more predictable and stable. You won't see a difference in the benchmarks as a result of this, but you'll probably find that your app's response times are more reliable.

Node.js was trying to be too aggressive, so it stopped doing that and allows V8 to run GC on its own schedule. 

> This leaves us with V8 GC-- which isn't guaranteed.

Indeterminate is not the same as "isn't guaranteed."

> Essentially, v8 was built by Google for Google Chrome. 

It's clear that you're not familiar with [the work](https://github.com/tc39/test262/pulls) I'm [involved in](https://github.com/rwaldron/tc39-notes). 

> Everybody else using v8 has to deal with issues like this one from time to time.

I should've been forthright with you in my first comment: I consulted directly with V8 implementation engineers before posting, to make sure that my generalized understanding of V8 memory management was both current and correct. I wouldn't have posted anything at all if I was wrong. We discussed your reply and that has informed my response, so please accept that I'm responding with correct information from the most credible source.

Zion-ICS: Thank you very much for your reply!
I need some time to ponder it.
One possible other root cause here is that the SWIG-generated node module, for whatever reason, doesn't seem to trigger v8 GC. I haven't proved it yet.
Our problem was that destructors weren't being called on exit without this cleanUp code. gdb seemed to confirm this behavior....

Propanu:
@rwaldron The issue is very likely related to the fact that we use SWIG to generate our V8 code and the node.js modules. Here's a quote from one of the swig generated wrappers:

	// Because there's no guarantee that v8 will ever call garbage collection,
	// we're adding a function that will allow a user to call it manually
	void cleanUp()
	{
		// Call the v8 garbage collector as long as there is memory to clean up
		// See https://codereview.chromium.org/412163003 for this API change
                #if (SWIG_V8_VERSION < 0x032838)
		while (!v8::V8::IdleNotification())
                #else
		while (!v8::Isolate::GetCurrent()->IdleNotification(1000))
                #endif
		{;}
	}

The memleak doesn't happen for all the drivers we write and it depends on the data structures we use. More so a problem when things get a bit more complex, i.e. arrays of fixed width integers.

@Zion-ICS Zion, do we need to set the variables to null in the examples, or is it enough to call the cleanup functions?

Zion-ICS:
gdb confirms that you need to set variables to null before calling cleanup.
This includes both class libraries and primitive wrappers such as uint8s, uint16s, etc.

Recognizing that nulling is needed in this environment, I've been nulling the upm node add-on library also. Even though the following line is probably a no-op:
`eldriver_lib = null;`

Propanu:
This was merged as fbbf56a49dd062eed043074ff14f9deb36eab41b. If you would like to continue this discussion about the V8 GC please use the issues section.

PR #130, opened 2015-03-09
groveO2: initial implementation

Jon-ICS:
This implements support for the Grove O2 (oxygen) sensor.


jrvandr:
To keep consistent with the naming of the rest of the modules in the repository, I've renamed this to groveo2 instead of groveO2.  Also, when the documentation was built with Doxygen, one of the flags it uses is set to disallow uppercase file and folder names and generated an error.
Thanks!  Merged in and closing.

PR #161, opened 2015-04-02
gp2y0a: renamed from gp2y0a21yk to indicate support for the gp2y0a famil...

Jon-ICS:
...y

Additionally, add some error checking in the constructor, and determine
the ADC resolution and use that in the voltage calculation rather than
a hardcoded 1024.0.


Propanu:
Merged in 77be6adf6de42d642c626b80b286359ae11423c1

PR #171, opened 2015-04-20
Refactor code, move to C++ and improve error handling

vvvverre:
This pull request starts by adding a .clang-format file, so that the code can be refactored into a common coding style easily. 
It then improves error handling inside the I2CLcd class and inside the Jhd1313m1 class, by throwing exceptions from the constructor if something goes wrong and by returning appropriate mraa_result_t values from methods. This way the user will be warned if the expected device type is not found on  the I2C bus or if the device does not respond. 
It also removes the I2C helper functions which are no longer required (it is not necessary to call mraa_i2c_address every transaction, as these functions were doing). 
The PR also moves the code from using the MRAA C API to the C++ API, which makes resource management easier. For example the mraa_i2c_context now automatically gets closed when the sensor object goes out of scope or gets deleted. 
It also replaces the #define statements inside the header files with const members of the namespace. This ensure that they are only available inside the lcd namespace, preventing potential clashes in the future.
Finally it adds the ssd.h header to the sources list in CMakeLists.txt, it seemed to be missing from there. 

In the future the extra error handling could be extended to the other LCD classes as well.

I have tested this code on all the screens, using an Edison.


arfoll:
Acked-by: Brendan Le Foll <brendan.le.foll@intel.com>

Propanu:
Thanks for the lcd improvements and for the clang format file! Running a clang format on the entire library is on our TODO list so this should be straightforward now.

Merged in 124a9eb41d81be9124b0f1d8cd2d63f676a43ea9

PR #173, opened 2015-04-23
adafruitms1438/CMakeLists.txt: fix building with BUILDSWIG=OFF

arfoll:
Missing if statement for building without SWIG (but not explicitely disabling
SWIGNODE & SWIGPYTHON)


Propanu:
Nice catch and I'm glad you're still following our code in such detail!
Merged in 9bb485367a613d6ed398874572c2ef703d429187

PR #177, opened 2015-04-24
sx6119: Initial implementation

Jon-ICS:
This was tested with the Grove FM Receiver.  Unfortunately, there is
no documentation on how to control the device using the D1 and D2 pins
that Seeed provides.  There is a switch on the device that can control
all of the elements, power, volume, and seek +/-.

The supplied example can turn the device on, and do a seek, but not
much else.


Propanu:
Merged, thanks!

PR #183, opened 2015-05-15
Fix the bug that the JS implementation of ST7735 class does not have its parent class's (GFX's) methods

thorikawa:
This pull request will fix the bug that the JS implementation of ST7735 class does not have its parent class's methods: e.g. GFX::fillScreen(uint16_t color).

Before this fix, when I run this source code:
```
var st7735 = require('jsupm_st7735');

var lcd = new st7735.ST7735(10, 7, 9, 8);
lcd.fillScreen(st7735.ST7735_RED);
```

I am seeing the following error message.
```
TypeError: Object #<ST7735> has no method 'fillScreen'
```

We need to include the parent class's header in the JS interface. Also I added the JS example code for testing it as well.

Thanks,

### DIFF
--- a/examples/javascript/st7735.js
+++ b/examples/javascript/st7735.js
@@ -1,0 +1,65 @@
+/*jslint node:true, vars:true, bitwise:true, unparam:true */
+/*jshint unused:true */
+/*
+* Author:  Takahiro Poly Horikawa <horikawa.takahiro@gmail.com>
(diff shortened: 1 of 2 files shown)


Propanu:
Great fix, thank you. Merged in 3d63b80c977f4657dbf73b80f85c26e17fb32549

PR #190, opened 2015-05-20
ad8232: Initial Implementation

Jon-ICS:
This driver implements support for the Sparkfun Single Lead Heart Rate
Monitor, based on the AD8232 chip.

It simply outputs ADC data that needs to be sent somewhere for
plotting.  The Sparkfun page has some suggestions.

Alternatively, if you have an oscilliscope that supports a 'Roll'
mode, you can get an EKG-like display by measuring the OUT pin.

https://www.sparkfun.com/products/12650


Propanu:
Thanks, merged here 517f6c1e88c699da3519e8e1a49f660cabffd0f7

PR #193, opened 2015-05-26
th02: fix non-working driver

Jon-ICS:
There were a few issues with this driver, corrected in this update.

1) i2c transactions were not working - I replaced the i2c
implementation with the MRAA i2c class implementation.

2) status check was inverted - fixed.

3) fixed up #defines in header file to avoid naming collisions (ADDR,
etc).

4) Added capability to supply bus and i2c address to ctor, setting
defaults of 0, and TH02_ADDR respectively.

NOTE: For proper operation on Edison using the arduino breakout
board, the voltage needs to be set to 3.3V rather than 5v.  On G2, 5v
works fine.

### DIFF
--- a/examples/c++/th02.cxx
+++ b/examples/c++/th02.cxx
@@ -50,7 +50,7 @@ main(int argc, char **argv)
 
     while (!doWork) {
         temperature = sensor->getTemperature ();
-        // humidity = sensor->getHumidity ();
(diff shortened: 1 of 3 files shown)


Propanu:
Great fix, thanks! Checked and merged in b633ecf97fa62e53348b96592a2d841af15f1678.

PR #201, opened 2015-06-02
Pull fixup uart for mraa7 changes

Jon-ICS:
A 'const' was added to the prototype declaration for mraa_uart_get_dev_path() in MRAA v7.  This causes all of the current UPM uart-based drivers to fail to build.

These patches fix up those issues, and UPM now builds with MRAA v7.


Propanu:
Thanks, these were already fixed here d6ce08227496b340ee22a9dbc710627c25d26022.

PR #209, opened 2015-06-16
at42qt1070: Additional functionality

vvvverre:
This patch series adds some additional functionality to the at42qt1070 touch sensor. 
It allows the user to read and change the low-power mode settings, the averaging factor and the adjacent key suppression groups.
It also allows the user to read the Chip ID register and does so in the constructor, as a sanity check.

Tested on Intel Edison on Xadow with a Xadow Q Touch sensor.


arfoll:
@Propanu can this get merged?

Propanu:
Merged, thanks a lot!

PR #212, opened 2015-06-29
cmake: Change error message when git not found to say UPM

arfoll:


Propanu:
I saw this before and wanted to fix it but got distracted it seems. Thanks!

PR #222, opened 2015-07-27
Pull lcd 1602 and rgb lcd (jhd1313m1) enhancments

Jon-ICS:
These patches add missing functionality to lcm1602, and make modifications so that it can be used as a base class for jhd1313m1 and others (in the future).

In addition, jhd1313m1 is modified to use lcm1602 as a base class (inheriting it's new functionality), and removing duplicate functionality already supported by the lcm1602 base class.

In addition, a sleep(5) is added to the jhd1313m1 example so that you have a chance to see the display, before it's destructor turns it off on exit.


Propanu:
Thanks a lot, added with 197b56003c85a7edb6095a46800fc24a0bad26de!

Issue #211, opened 2015-06-27
adafruitss.cxx frequency is hard coded


wolfmanjm:
It appears that the PWM frequency is hard coded here...
https://github.com/intel-iot-devkit/upm/blob/master/src/adafruitss/adafruitss.cxx#L60

Instead of using the passed in frequency. This means that although the frequency may be changed the pulse width will be incorrect for the given frequency.

I'll fix this and test it before submitting a PR for the fix.

Propanu:
You might want to take a look at the PCA9685 driver provided by @Jon-ICS. He used it to implement the Adafruit MotorShield, since that one also uses the same PWM driver and might come in handy here.

Jon-ICS:
This driver should probably be re-written to use pca9685 :)

wolfmanjm:
this particular bug is not related to the PCA9685, it is specific to determining the pulse width for a servo used later on.

I agree using the PCA9685 would make sense though to avoid duplication, but that is simply a refactor :)

wolfmanjm:
I've fixed and tested the pwm frequency bug. I'm still looking at how hard it is to use the PCA9685 driver. It initially looks like there are some missing features, but I need to investigate further.

Propanu:
Can you please submit a pull request with your fix?

PR #223, opened 2015-08-01
Fix the period setting in setPWMFreq()

wolfmanjm:
Tweak the fudge factor so it is more accurate for 50Hz and 60Hz periods. (Tested on Logic 4 and it is spot on now)
Get rid of compiler warnings when -Wall is set
Allow for float angle to be passed into the servo() method for more accurate positioning

Fixes issue #211


Propanu:
Thanks a lot. Merged in 7260c78c335981f1190eb166373b55f9102588d4!

PR #227, opened 2015-08-06
eboled: Initial implementation

Jon-ICS:
This driver supports the Sparkfun 64x48 pixel OLED Edison block:

https://www.sparkfun.com/products/13035

It is based on an ssd1306, but with some modifications (custom COM pin
mapping and a custom column offset).  It uses SPI to communicate, and
since it is an Edison Block, you don't really have any options for
different bus and pin assignments.


Propanu:
Thank you! Merged in 55e807698887f9322c529e8d12e31f1fb6a9b912

PR #229, opened 2015-08-07
Pull upm exceptions

Jon-ICS:
upm: add facility for exception handling in upm
    
We add a new src/upm_exception.i interface file for SWIG to catch
common exceptions and propagate them through SWIG.
    
src/upm.i is modified to include this interface file, so all UPM
drivers have it.
    
In theory, this should be language agnostic - if the target language
supports exceptions, then it should just work.

Next, existing drivers with their own exception classes are modified to remove those classes and use appropriate standard exceptions instead.

### DIFF
--- a/src/ecs1030/ecs1030.cxx
+++ b/src/ecs1030/ecs1030.cxx
@@ -25,22 +25,18 @@
 #include <iostream>
 #include <unistd.h>
 #include <stdlib.h>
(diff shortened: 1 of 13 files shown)


Propanu:
Awesome work, thanks! 018e09e5b2251109e4036afbb27532eaffb85cb2

PR #233, opened 2015-08-17
More fixes regarding Java bindings and packaging

andreivasiliu2211:


Propanu:
The forced redefinition of SWIGJAVA f072499 was left out as it doesn't seem to be required and breaks the build.
Also, eab0a6f & 432f9c4 cannot be merged until MRAA adds your *types.hpp* extensions.
Everything else was merged and I will leave this open so that we can pull in the last changes once the new header gets added.

arfoll:
new headers are now added, sorry about the delay guys, getting the ft4222 stuff merged took a long time!

Propanu:
Are you absolutely sure f072499 is required for callbacks? Everything else was merged and the build seems flawless. Try my https://github.com/Propanu/upm/tree/java branch to confirm, otherwise open up a PR with the other fix we discussed. Will get merged on master once MRAA fix is in.

andreivasiliu2211:
We managed to make the callbacks work using another define, JAVA_CALLBACK,
and now the warnings are gone. Ștefan (cc'd) tried that on Friday and he
says it works.

Andrei

Propanu:
This should be fully merged now!

PR #242, opened 2015-08-26
Eboled driver improvements

tylergibson:
Adding functionality to the Sparkfun Edison Blocks OLED display.

Changed the SPI write strategy to use purely horizontal mode with word (uint16) writing.  Dramatically improves write speed while still avoiding the known SPI crash issue on Edison with MRAA and SPI.write with large buffers.

Fixed some errors in the horizontal addressing mode, moved to using a single status framebuffer and ported the font drawing functions from /src/st7735/gfx.cxx.

Implemented the basic drawing functionality based on Adafruit's GFX library and updated the node.js example to be much more complete.  Adding the appropriate licensing text and attributions.


tylergibson:
Going to squash my commits and do this correctly in another pull request :)

Propanu:
Thanks a lot that would really help!

PR #252, opened 2015-09-09
Pull kit sensors throw exceptions on fatal errors

Jon-ICS:
This PR adds code to the sensors to throw exceptions on fatal errors, errors that could not otherwise be differentiated from normal operation, and various range and operating errors.

There are still error conditions that can occur that will output error messages without throwing exceptions.  Maybe these can be addressed in the future if that behavior is desired.


Propanu:
Awesome contribution, merged with some changes. Thanks a lot for this!

PR #253, opened 2015-09-14
Pull gas convert to mraa c++ aio api and add mq4 6 7 8

Jon-ICS:
This PR changes the gas sensors to use the MRAA C++ Aio interface. 

In addition, it adds support for the DFRobot MQ4, MQ6, MQ7, and MQ8 gas sensors.


Propanu:
Merged, thanks!

PR #258, opened 2015-09-16
grovemd: add support for 'mode1' stepping and add stepper examples

Jon-ICS:
The current grovemd driver supported 'mode2' stepping, where the
driver simply passed various stepper commands to the board for it to
carry out on it's own.

This doesn't work very well (or at all if you have old/buggy firmware)
so add a new 'mode1' stepper capability.  This mode lets the driver
manually control the stepping operation without requiring special
firmware.

This is now the default and recommended mode to use for stepper motors
on this device.  It is also more flexible in terms of the maximum
number of steps you can do (mode2 was limited to 254 steps max).

This was tested using a bipolar NEMA-17 stepper motor with an
external 12v power supply.

Note: 'Mode1' and 'Mode2' are the Seeed Studio terms for these different
stepping modes.


Propanu:
Great fix, thanks 83e62aabba3c069f232a6110522a5d3f5f2fcc7e!

PR #261, opened 2015-09-21
docs: C++ to Java mapping of sample files for documentation

mircea:
Signed-off-by: Mircea Bardac <mircea.bardac@intel.com>


Propanu:
Added, thanks!

PR #268, opened 2015-09-23
xbee: Initial implementation

jontrulson:
This is a basic serial module that allows access to various XBee
devices via a UART port.  It was tested with the XBee S6B WiFi Module
and the XBee S1 802.14.4 module.


malikabhi05:

PR #271, opened 2015-09-25
urm37: Initial implementation

jontrulson:
This driver was tested with the DFRobot URM37 Ultrasonic Ranger, V4.
Both UART and analog access modes are supported.


malikabhi05:

PR #274, opened 2015-09-28
adxrs610: Initial implementation

jontrulson:
This module was developed and tested on a ADXRS610 Gyro Beakout
board.

http://www.dfrobot.com/index.php?route=product/product&product_id=642


malikabhi05:

PR #275, opened 2015-09-29
java: Enabled directors feature where needed

andreivasiliu2211:
Directors feature is disabled by default. Without this commit, the SWIG interfaces containing callbacks that do not have directors="1" will not work.


Propanu:
Merged, thanks!

PR #277, opened 2015-09-29
docs: Sync sensor library names in documentation with the actual libr…

mircea:
…ary names

Signed-off-by: Mircea Bardac <mircea.bardac@intel.com>


arfoll:
Merged

PR #278, opened 2015-09-30
Character LCD modules - support for various configurations

skiselev:
It appears that currently UPM (lcm1602) supports only 20x4 modules. This patch adds support for other LCD configurations 8x1, 16x1, 16x2, 20x2, 40x2, 16x4, 20x4.
It also adds an example for using an LCD with parallel interface.

As a side note lcm1602 is not an obvious name for character LCD modules. Consider renaming it. hd44780 or just 'lcd' are probably some better choices.

### DIFF
--- a/src/lcd/lcm1602.cxx
+++ b/src/lcd/lcm1602.cxx
@@ -213,7 +213,7 @@ Lcm1602::setCursor(int row, int column)
             if (m_numColumns > 8)
             {
                 offset = (column % (m_numColumns / 2)) +
-                         (column / (m_numColumns / 2)) * 0x40;
+                         (column / (m_numColumns / 2) * 0x40;
             }
             break;
         case 2:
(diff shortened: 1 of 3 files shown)


src/lcd/lcm1602.cxx
```diff
@@ -196,9 +200,49 @@ mraa::Result
...
+            break;
+        case 2:
+            // this should work for 40x2, 20x2, and 16x2 displays
+            // DDRAM mapping:
+            // 00 .. 0F .. 13 .. 27
+            // 40 .. 4F .. 53 .. 67
+            offset += row * 0x40;
+            break;
+        case 4:
+            if (m_numRows == 16)
```
Propanu: Doesn't this have to be m_numColumns?

skiselev: Yes it does :-)

skiselev:
m_numColumns typo fixed. Added more flexible support for single row displays.

Propanu:
Merged, thank you!

PR #283, opened 2015-10-02
Pull remove setttyfd compiler warnings

jontrulson:
These commits remove an unused setTtyFd() method in the older UART based UPM sensors that caused compiler warnings.


Propanu:
Merged, thanks!

PR #289, opened 2015-10-10
lsm303: Fixing an incorrect accelerometer address

xbolshe:
Signed-off-by: Eugene Bolshakov <pub@relvarsoft.com>

I2C address was incorrect.
It shall be 0x19, as descibed in a datasheet: http://www.st.com/st-web-ui/static/active/en/resource/technical/document/datasheet/DM00027543.pdf

### DIFF
--- a/src/lsm303/lsm303.h
+++ b/src/lsm303/lsm303.h
@@ -33,8 +33,8 @@
 namespace upm {
 
 /* LSM303 Address definitions */
-#define LSM303_MAG 0x1E // assuming SA0 grounded
-#define LSM303_ACC 0x18 // assuming SA0 grounded
+#define LSM303_MAG 0x1E
+#define LSM303_ACC 0x19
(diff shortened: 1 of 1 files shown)


xbolshe:
LSM303D has SA0 pin to select I2C less significant bit of the device address.
http://www.st.com/web/en/resource/technical/document/datasheet/DM00057547.pdf

But LSM303DLHC has not.

Moreover, Adafruit LSM303 board uses LSM303DLHC.
https://www.adafruit.com/products/1120

WaveShare LSM303 board that I use has also LSM303DLHC.
http://www.wvshare.com/product/LSM303DLHC-Board.htm

So, looking on LSM303 products on a market it is better to select an accelerometer address as 0x19 by default.

malikabhi05:
Implemented and pushed upstream (intel-iot-devkit)

PR #292, opened 2015-10-13
Cleaned up code.

MarcGraham:
Removed some unneeded calls in update function. Changed call in
micsv89.js sample file to use update instead of start.

Tested with UPM. 4.0. The micsv89 is sensitive to other devices pulling
the I2c bus up, so it is not likely to work on the Arduino breakout.
I2c 1 on mini breakout works fine.

—Signed-off-by Marc Graham <marc@m2ag.net>


malikabhi05:
Pushed upstream (intel-iot-devkit)

PR #293, opened 2015-10-13
Fix the example for SparkFun Block for Intel Edison - OLED

skiselev:


Propanu:
Should be all there now dc97adf57878c2fdcb8699fc675f27be5e7f6188. Other example was renamed. Thanks!

PR #294, opened 2015-10-13
dfrph: Initial implementation

jontrulson:
The driver implements support for the DFRobot pH sensors.  It was
tested with both the standard and Pro versions, calibrated with
standard buffer solutions at pH 4.01 and pH 7.0.


malikabhi05:

PR #299, opened 2015-10-27
ozw: Initial implementation

jontrulson:
This module was developed using the OpenZWave library (v1.3):

http://www.openzwave.com/

It was developed using a collection of devices (switches and a
multi-sensor) connected via an Aeon Z-Stick Gen5 USB dongle.

It can be used to query (and, where appropriate, set) Values on Nodes
connected to the ZWave network.

Checks are made in src/ozw/CMakeLists.txt to ensure that the
libopenzwave library is installed (via pkg-config).  If not present,
then neither the module, nor the example will be built.


Propanu:
Need to decide whether we want this in MRAA or in UPM with sensor driver targets depending on this (similar to pca9685 shields). I can see both options working. @arfoll any ideas?

jontrulson:
Uhm, what did I miss?  where did all these commits come from?

Propanu:
Thanks a lot, merged here fd509c7d790991c6ce65c4613d3ab92de71e5113!

PR #301, opened 2015-10-29
use cmake libdir

tripzero:
Using cmake's libdir variable instead of constructing our own helps to avoid issues related to building upm on multilib enabled systems.


Propanu:
Thank you! Merged as 6e2355ce9afc98cbeb0c7c2edca25cf471f8a6fe.

PR #305, opened 2015-11-17
sx1276: Initial implementation

jontrulson:
This driver was developed using an SX1276 based shield on the Galileo
G2.  It requires 3.3v of operation.  It does not work with Edison, due
to SPI issues.


Propanu:
Thanks, merged with 2a1af1399a09e36f20dfe3f7775a462d897f9d8c!

PR #307, opened 2015-11-20
Pass reference to current object to ISR.

jjrob13:
This will allow users to call ackEdgeDetected() on the object passed into the function pointer, instead of requiring them to have a global reference to the object.

### DIFF
--- a/examples/c++/hcsr04.cxx
+++ b/examples/c++/hcsr04.cxx
@@ -24,7 +24,7 @@
 
 #include <unistd.h>
 #include <iostream>
-#include <upm/hcsr04.h>
+#include "hcsr04.h"
 #include <signal.h>
 #include <stdlib.h>
 #include <sys/time.h>
--- a/examples/c++/hcsr04.cxx
+++ b/examples/c++/hcsr04.cxx
@@ -24,12 +24,10 @@
 
 #include <unistd.h>
 #include <iostream>
-#include "hcsr04.h"
+#include <upm/hcsr04.h>
 #include <signal.h>
 #include <stdlib.h>
 #include <sys/time.h>
-
-upm::HCSR04 *sonar = NULL;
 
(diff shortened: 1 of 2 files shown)


Propanu:
Thanks for the contribution, can you please sign-off your commit?
The fix is good, the example will have to be updated to match this. Personally, I don't think the ultrasonic sensors need to expose more than _getDistance()_ as a public function.

jjrob13:
Thanks for the quick response.  Yeah, I thought this would make things
easier after spending a few hours trying to pass binded lamdba functions as
c function pointers to no avail.

Can you clarify on what you mean by "sign-off your commit"?

Thanks,

JJ Robertson

Propanu:
Do a quick amend on the commit and add your signature line with the --sign-off (-s) option.
Example:

`git commit --amend --sign-off --no-edit`

Thanks!

P.S. You can find our contribution guidelines and code signing requirements [here](https://github.com/intel-iot-devkit/upm/blob/master/docs/contributions.md#code-signing).

jjrob13:
Thanks for your help.  I amended the commit and signed off.

jjrob13:
I updated the hcsr04.cxx example file to illustrate the new usage, too.  There is also an additional commit with me updating an include statement to what it initially was.  If this is the only way that the interrupt function will be used, it might make sense to abstract that away and just provide a default implementation.  I can do that, too, if you think it is worthwhile.  It would bring it closer to just exposing the getDistance() function, as you mentioned earlier.

Propanu:
I've merged your initial change but turns out that the new model doesn't work as expected with our Java bindings. Thus we decided to hide the interrupt function internally and only expose getDistance(). This should simplify the sensor's API greatly for end users.

Thank you for your contribution though and I do hope the new use model will be suitable for your application too!

PR #314, opened 2015-12-09
nlgpio16: Initial implementation

jontrulson:
This is a USB device from Numato Labs that is accessed via a UART.  It
provides 16 GPIO's, 7 of which can be configured as analog inputs.


Propanu:
Thanks, merged!

PR #315, opened 2015-12-10
cmake: Use include/lib paths from MRAA pkgconfig for cmake

pylbert:
If successful, the MRAA pkgconfig sets MRAA_INCLUDE_DIRS and MRAA_LIBDIR.
These can be used by dependent cmake files.

    * Updated src CMakeLists.txt to use MRAA pkgconfig locations in
      upm_module_init
    * Updated c++ example CMakeLists.txt to use MRAA pkgconfig locations
    * Updated UPM building.md to reflect the change (removed the need to set
      CPLUS_INCLUDE_PATH and LIBRARY_PATH

### DIFF
--- a/docs/building.md
+++ b/docs/building.md
@@ -5,12 +5,10 @@ UPM uses cmake in order to make compilation relatively painless. Cmake runs
 build out of tree so the recommended way is to clone from git and make a build/
 directory.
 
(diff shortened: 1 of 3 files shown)


Propanu:
Merged into master, thanks!

PR #316, opened 2015-12-11
java: added auto build java examples in cmake

stefan-andritoiu:
Signed-off-by: Stefan Andritoiu <stefan.andritoiu@intel.com>


Propanu:
Merged. The short casts look inconvenient, thus it might help to typemap all the uint*_t types to the corresponding unsigned int, similar to what was done here 387b2b86b02ee1663920969afa3cf9ed47296e03. Thanks!

PR #320, opened 2015-12-17
cmake: added doxygen doc content to python sensor modules

pylbert:
Include doxygen documentation collateral in python sensor modules.  This runs
when BUILDDOC and BUILDSWIGPYTHON are set.  The general flow is:

    doxygen -> index.xml -> doxy2swig.py -> pyupm_doxy2swig.i -> swig

    * Removed registered trademark symbol from README.ccp.md since this
    does not agree with the include version of doxy2swig.py.

    * Removed per-sensor target run of doxy2swig.py.  pyupm_XXX targets now
    depend on the output of doxy2swig.

    * Added custom command to build pyupm_doxy2swig.i from index.xml.  If
    BUILDDOC=OFF, an empty stub pyupm_doxy2swig.i is used.  A single output
    .i file is generated which is used by ALL pyupm_XXX sensor targets.

    * Added %include pyupm_doxy2swig.i per pyupm_XXX sensor target.

    * Added swig source file property which includes build directory.

### DIFF
--- a/doxy/README.cpp.md
+++ b/doxy/README.cpp.md
@@ -30,7 +30,7 @@ Multi-sensor samples for the starter and specialized kits can be found in the
 
 Supported [sensor list](http://iotdk.intel.com/docs/master/upm/modules.html) from API documentation.
 
-You can also refer to the [Intel® IoT Developer Zone](https://software.intel.com/iot/sensors).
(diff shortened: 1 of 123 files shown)


Propanu:
Merged into master, thanks!

PR #326, opened 2015-12-29
sx1276 and adxrs610: C++11 incompatibility fixes

xbolshe:
Signed-off-by: Eugene Bolshakov <pub@relvarsoft.com>


Propanu:
Merged, should make it into upstream shortly! Thanks a lot for the patch, hope the libs build nicely with nodejs4+ now.

PR #333, opened 2016-01-13
hdxxvxta: Initial implementation

jontrulson:
This driver was developed using the Veris HD2NVSTA1 humidity
transmitter.  The 'T' variant supports a temperature transmitter as
well.  Both signals are provided by the device as analog 0-5Vdc or
0-10Vdc outputs.

The A1 variant supports a temperature range of -40C-50C, while the A2
variant supports a range of 0C-50C.  Humidity ranges for all devices
in this device family range from 0% to 100% (non-condensing).

Temperature measurement can be disabled by passing -1 as the
temperature analog pin to the constructor.


Propanu:
Merged, thanks!

PR #334, opened 2016-01-16
doc: Added ds1307 to knownlimitations list

pylbert:
The Grove RTC (DS1307) does not work on the Intel Edison Arduino
board but will work with the mini-breakout.  Updated the list of
known limitations in docs.  Updated the ADXL345 to show 16g, and
fixed one line which went over 80 chars.


Propanu:
Merged, thanks!

PR #336, opened 2016-01-20
rhusb: Initial implementation

jontrulson:
The driver implements support for the Omega RH-USB Humidity Probe with
Temperature sensor.  It connects via an integrated USB cable, and
appears as a serial port.

It does not currently work with Edison (as of this date) due to
missing ftdi_sio and usbserial kernel support.

It was implemented and tested on the Galileo 2.


Propanu:
Merged, thanks!

PR #328, opened 2016-01-04
Migrate IsrCallbacks to Java's Runnable class

petreeftime:
Signed-off-by: Petre Eftime <petre.p.eftime@intel.com>

Depends on: https://github.com/intel-iot-devkit/mraa/pull/397

PR #339, opened 2016-01-28
t3311: Added libmodbus linkage for Java build

andreivasiliu2211:


src/t3311/CMakeLists.txt
```diff
@@ -17,5 +17,8 @@ if (MODBUS_FOUND)
     if (BUILDSWIGPYTHON)
       swig_link_libraries (pyupm_${libname} ${MODBUS_LIBRARIES} ${PYTHON_LIBRARIES} ${MRAA_LIBRARIES})
     endif()
+    if (BUILDSWIGJAVA)
+      swig_link_libraries (javaupm_${libname} ${MODBUS_LIBRARIES} ${MRAA_LIBRARIES} ${JAVA_LIBRARIES})
```
Propanu: @petreeftime updated the linker flags in #328, shouldn't they be `${MRAAJAVA_LDFLAGS} ${JAVA_LDFLAGS}` now?

petreeftime:
Yes, you should link with MRAAJAVA instead of MRAA. I think _LIBRARIES should be fine, since libmraajava.so has a shortcut in /usr/lib, but it needs to be checked.

Propanu:
Merged, thanks!

PR #340, opened 2016-01-28
ds18b20: Initial implementation

jontrulson:
This driver supports, and was tested with, a DS18B20 1-wire
Temperature Sensor using external power.

This device requires the use of a UART to provide access to a Dallas
1-wire bus, via a new facility supported by MRAA (once the relevant PR
is accepted), using the UartOW access class.  It is important to
realize that the UART is only being used to access and control a
Dallas 1-wire compliant bus, it is not actually a UART device.

Multiple DS18B20 devices can be connected to this bus.  This module
will identify all such devices connected, and allow you to access them
using an index starting at 0.

Parasitic power is not currently supported due to the very tight 10us
limit on switching a GPIO properly to supply power during certain
operations.  For this reason, you should use external power for your
sensors.

Setting the alarm values (Tl, Th) is also not supported, since this is
only useful when doing a 1-wire device search looking for devices in
an alarm state, a capability not yet supported in MRAA.  In reality,
this is trivial to handle yourself in your application.

### DIFF
--- a/examples/c++/CMakeLists.txt
+++ b/examples/c++/CMakeLists.txt
@@ -258,6 +258,7 @@ add_example (smartdrive)
 if (HAVE_FIRMATA)
   add_example (curieimu)
 endif ()
+add_example (ds18b20)
 
 # These are special cases where you specify example binary, source file and module(s)
 include_directories (${PROJECT_SOURCE_DIR}/src)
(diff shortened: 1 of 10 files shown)


Propanu:
Will merge as soon as MRAA adds #415 and tags.

pylbert:
Thanks Jon, I updated the header filename and merged.

PR #341, opened 2016-01-30
my9221: rewrite to include new functionality and serve as a base class

jontrulson:
This driver has been rewritten to support some new functionality, be
    more generic and fix up some old bugs.  Multiple MY9221's can now be
    chained together and are supported by the MY9221 base class.
    
    In addition, the Grove LED Bar and Grove Circular LED drivers have
    been incorporated into the my9221 library, using the new MY9221 class
    as their base class.  Examples have been fixed to work with the new
    library, and renamed where needed.
    
    The current grovecircularled driver has been removed as it is now a
    part of the my9221 library.
    

Propanu:
Great update, merged into 0.5.1! Thanks a lot.

PR #353, opened 2016-02-20
cwlsxxa: Initial implementation

jontrulson:
The driver was developed using the Veris CWLSHTA CO2 Gas sensor.  The
'T' variant supports a temperature sensor, and the 'H' variant
supports a humidity sensor.

All 3 signals are provided by the device as analog 0-5Vdc, 0-10Vdc, or
4-20ma loop current outputs.  For devices supporting temperature, the
valid temperature range is 10C to 50C.  The humidity ranges from 0% to
100% (non-condensing).  The CO2 sensor ranges from 0 to 2000 ppm.

This driver used the 5Vdc outputs for obvious reasons.  Your MCU must
be configured for 5V operation.  In addition, you must configure the
sensor (via it's configuration switches) to output 0-5VDC only.  Using
any other analog reference voltage will require the appropriate
external circuitry (such as a voltage divider) in order to interface
safely with your MCU.

### DIFF
--- a/examples/c++/CMakeLists.txt
+++ b/examples/c++/CMakeLists.txt
@@ -239,6 +239,7 @@ add_example (hdxxvxta)
 add_example (rhusb)
 add_example (apds9930)
 add_example (kxcjk1013)
+add_example (cwlsxxa)
 
 # These are special cases where you specify example binary, source file and module(s)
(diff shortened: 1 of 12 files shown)


Propanu:
Merged, thanks!

PR #355, opened 2016-02-23
apa102: Initial Contribution

yannickadam:
APA102 LED Strips provide individually controllable LEDs through a SPI interface.
For each LED, brightness (0-31) and RGB (0-255) values can be set.

Tested on Intel Edison

Signed-off-by: Yannick Adam <yannick.adam@gmail.com>


yannickadam:
The driver needs faster writes for the whole strip. Will resubmit once done.

PR #358, opened 2016-02-27
teams: Initial implementation

jontrulson:
The driver adds support for the Veris TEAMS Temperature Transmitter.

It provides it's output via a 4-20ma current loop.  The supported
temperature range is 10C to 35C.

This sensor was developed with a Cooking Hacks (Libelium)
4-channel 4-20ma Arduino interface shield.  For this interface,
the receiver resistance (rResistor) was specified as 165.0
ohms.


Propanu:
Merged, thanks!

PR #360, opened 2016-03-01
APA102: Initial implementation

yannickadam:
APA102 LED Strip driver

These LEDs (also called DotStar by Adafruit) have an integrated controller. It is SPI based, and each LED can be controlled individually in the strip.

For each LED, it is possible to provide values for brightness (0-31) and red, green, blue (0-255) components. 

This driver provides utility methods to address a single, a range, or all leds in a strip.

It has been created and tested on an Intel Edison, using a custom built image based on the latest kernel source at 01org/edison-linux (this includes fixes to SPI).

Signed-off-by: Yannick Adam <yannick.adam@gmail.com>


Propanu:
Great contribution thanks a lot for submitting this! Merged.

PR #362, opened 2016-03-02
tex00: Initial Implementation

jontrulson:
The Veris TEX00 temperature sensor family is made up of a series of
RTD thermistors in wall mount packaging.

This driver was developed using the TED00, which utilizes a 10K Ohm
Type 2 thermistor.  However, this driver can support the other 12
variants of the TE series as well by providing the correct sensor type
to the class constructor.  These other sensor types have not been
tested.  Only the TED00 hardware was tested with this driver.

This sensor must be connected as part of a voltage divider, with the
balancing resistor ideally matched to the sensor's 25C detection
range.  For the TED00 (10kt2), a 10K Ohm (1% tolerance) resistor was
used in a circuit like the following:

GND o----|TED00(10k2)|----o----|balanceResistor(10K)|----o VCC (+5vdc)
                          |
                          |
                          |----o A0 (analog input to MCU)

A 3.3vdc voltage can be used as well if desired.

### DIFF
--- a/examples/c++/CMakeLists.txt
+++ b/examples/c++/CMakeLists.txt
@@ -239,6 +239,7 @@ add_example (hdxxvxta)
 add_example (rhusb)
 add_example (apds9930)
 add_example (kxcjk1013)
+add_example (tex00)
 
 # These are special cases where you specify example binary, source file and module(s)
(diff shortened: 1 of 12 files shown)


Propanu:
Merged, thanks!

PR #363, opened 2016-03-03
cmake: Enable C++11 standards support

jontrulson:
This patch checks for, and enables C++11 support for building UPM.
This should work for all cmake versions currently supported by UPM
(2.8.11+), and any compiler (clang/gcc) that was released in this
decade.

Support can be specifically disabled by passing '-DENABLECXX11=OFF' to
cmake, though modules requiring this support will not build.

C++11 support is enabled by default.


jontrulson:
Ok, so on my system, these narrowing conversions just generate warnings, but on CI they appear to be errors.  I will close this PR and submit a new one that also addresses these issues.  Thanks.

PR #364, opened 2016-03-03
Enable C++11 standards support, and fix narrowing conversions

jontrulson:
The first patch modifies the main CMakeLists.txt file so that it checks for, and enables C++11 support for building UPM.

This should work for all cmake versions currently supported by UPM
(2.8.11+), and any compiler (clang/gcc) that was released in this
decade.
    
Support can be specifically disabled by passing '-DENABLECXX11=OFF' to
cmake, though modules requiring this support will not build.
    
C++11 support is enabled by default.

The other patches in this PR fix narrowing conversions that were tolerated in pre-C++11, but are warnings or errors in C++11.  These fixes are still valid even if not building with C++11 support.


Propanu:
All merged, thanks!

PR #368, opened 2016-03-13
mics6814: initial implementation

g-vidal:
migration from Jacky Zhang work Seeed Technology

Signed-off-by: g-vidal gerard.vidal@ens-lyon.fr


g-vidal:
Is there something wrong with this pull request? Do I have to do something more?
Thanks

Propanu:
Hi,

Mics6814 should be ready to go although I would really prefer to have code comments changed to doxygen style. Also, the copyright header can be simplified to include both you and Jacky as authors followed by the MIT license similar to what we have for other drivers.

MQ135 is a different story though:
* The license link helps but unfortunately our libraries are also released as binary, not only source. We prefer not to impose a restrictive license on our users.
* PPM conversion seems to be off as other stated, the resistor value computation is useless (and likely wrong) as you only need the Rs/R0 ratio. Even the author listed so in his comments.
* Author didn't acknowledge D. Gironi, likely the place where he got the code from initially: http://davidegironi.blogspot.de/2014/01/cheap-co2-meter-using-mq135-sensor-with.html.
* Davide got the power function approximation correct, at least for NH4, as can be seen from his work: 0.9995 CoD. For CO2 however, exp fits better with 0.994 vs. 0.979. Pow also adds more error as you increase in ppm which is something you'll likely want to avoid.
* I can't use the image file provided for this sensor unless you own it.

I would use an exponential approximation instead with a new implementation that you can call your own and that will also solve the licensing issue. The only thing from GeorgK's code that I might consider is the correction for temperature/humidity, but didn't check against datasheet for accuracy. Also offering to help out with it if needed!

g-vidal:
Le 24/03/2016 22:20, Mihai Tudor Panu a écrit :
>
> Hi,
>
> Mics6814 should be ready to go although I would really prefer to have
> code comments changed to doxygen style.
>
I don't know how to do it  I will reada little bit and do it
>
> Also, the copyright header can be simplified to include both you and
> Jacky as authors followed by the MIT license similar to what we have
> for other drivers.
>
OK I'll do that
>
> MQ135 is a different story though:
>
This is more work and maybe with few chances to reach a solution.
I'll check later

thanks for the answer
>
>   * The license link helps but unfortunately our libraries are also
>     released as binary, not only source. We prefer not to impose a
>     restrictive license on our users.
>   * PPM conversion seems to be off as other stated, the resistor value
>     computation is useless (and likely wrong) as you only need the
>     Rs/R0 ratio. Even the author listed so in his comments.
>   * Author didn't acknowledge D. Gironi, likely the place where he got
>     the code from initially:
>     http://davidegironi.blogspot.de/2014/01/cheap-co2-meter-using-mq135-sensor-with.html.
>   * Davide got the power function approximation correct, at least for
>     NH4, as can be seen from his work: 0.9995 CoD. For CO2 however,
>     exp fits better with 0.994 vs. 0.979. Pow also adds more error as
>     you increase in ppm which is something you'll likely want to avoid.
>   * I can't use the image file provided for this sensor unless you own it.
>
> I would use an exponential approximation instead with a new
> implementation that you can call your own and that will also solve the
> licensing issue. The only thing from GeorgK's code that I might
> consider is the correction for temperature/humidity, but didn't check
> against datasheet for accuracy. Also offering to help out with it if
> needed!
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/intel-iot-devkit/upm/pull/368#issuecomment-201026451>
>

-- 
Gérard Vidal
Chargé de mission normes et veille technologique et numériques auprès du
Directeur de l'IFÉ.
Université de Lyon : IFÉ / ENS de Lyon 15 Parvis rené Descartes 69342
Lyon Cedex07.
tel : [+33] (0)4 26 73 12 60.
onglet logos

Propanu:
>> Hi,
>>
>> Mics6814 should be ready to go although I would really prefer to have
>> code comments changed to doxygen style.

> I don't know how to do it  I will reada little bit and do it

I can modify the comment blocks if you update the header license for mics6814. Here's a good [example](https://github.com/intel-iot-devkit/upm/blob/master/src/stepmotor/stepmotor.h).

PR #373, opened 2016-03-17
Implement CurieIMU sensor for MRAA Firmata client

deadprogram:
This PR implements a sensor to communicate with the built-in devices that are in the Arduino101/Genuino101. It uses MRAA acting as a Firmata client. 

This sensor requires that the FirmataCurieIMU plugin to ConfigurableFirmata be compiled and installed on a Arduino101/Genuino101 board.

Here is a very simple example of how it can be used:

```c
#include <unistd.h>
#include "curieimu.hpp"
#include "mraa.h"
#include "mraa/firmata.h"

int
main(int argc, char **argv)
{
    mraa_init();
    mraa_add_subplatform(MRAA_GENERIC_FIRMATA, "/dev/ttyACM0");

    upm::CurieImu* sensor = new upm::CurieImu();

    int x, y, z;
    sensor->readAccelerometer(&x, &y, &z);
    printf("accelerometer is: %d, %d, %d\n", x, y, z);

   delete sensor;
}
```

### DIFF
--- a/examples/c++/CMakeLists.txt
+++ b/examples/c++/CMakeLists.txt
@@ -240,6 +240,7 @@ add_example (rhusb)
 add_example (apds9930)
 add_example (kxcjk1013)
 add_example (ssd1351)
+add_example (curieimu)
 
 # These are special cases where you specify example binary, source file and module(s)
 include_directories (${PROJECT_SOURCE_DIR}/src)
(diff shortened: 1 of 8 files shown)


Propanu:
Thanks @deadprogram! @arfoll will add a cmake file for this that checks for the MRAA-Firmata symbols when building. Looks like swig interfaces will also need to spiced up with the pointer typemaps. I'll merge once CMakeLists.txt is ready.

arfoll:
merged in curieimu

PR #380, opened 2016-03-29
ads1x15 fix

whbruce:
cmake update to avoid conflicts with pre-installed version of UPM and ads1x15 examples build fix.


Propanu:
This was merged, thanks!

PR #382, opened 2016-04-04
ozw: Removed add_dependencies call from ozw CMakeLists

pylbert:
Small change to get rid of a warning in newer cmake versions.

Versions of cmake (>= 3.0) throw a warning on the add_dependecy
method for non-existant dependencies (add_dependency call before
target_link_libraries call).

Removed the call to add_dependency since target_link_libraries should
provide the same functionality.


Propanu:
Should be extended to Modbus drivers too, since they also link externally.

Propanu:
Merged!

PR #383, opened 2016-04-05
cmake: Solved issue with nodejs installation path

andreivasiliu2211:
Changed ${LIB_INSTALL_DIR} with lib, because the variable expands to
/usr/lib, making the install path /usr/usr/lib/node_modules, which
is incorrect. Now the install path is /usr/lib/node_modules.


Propanu:
Thanks, merged!

PR #393, opened 2016-04-21
kxcjk1013: code cleanup and update sensor description

laykuanloon:
Code cleanup and proper close iio device in destructor.

KXCJK-1013 is tri-axis accelerometer from Kionix.
This sensor can measure acceleration in metre per second squared or in
G-forces.

The library provided is libupm-kxcjk1013.so.0.4.0.
The example provided is kxcjk1013-example where it will print x,y,z axis when
trigger buffer data is ready.

Signed-off-by: Lay, Kuan Loon <kuan.loon.lay@intel.com>


pylbert:
Merged.  Thanks Kuan Loon.

PR #394, opened 2016-04-21
Apds9930 proper close iio

laykuanloon:
apds9930: proper close iio device in destructor and update sensor description

    Proper close iio device in destructor.

    APDS-9930 is light and proximity sensor from Avago Technologies.
    This sensor can measure ambient light in lux and proximity distance in 100mm.

    The library provided is libupm-apds9930.so.0.4.0.
    The example provided is apds9930-example where this app will print the value of
    ambient light and proximity in every 1 second.


pylbert:
Merged.  Thanks Kuan Loon.

PR #395, opened 2016-04-21
Enable l3gd20

laykuanloon:
L3GD20: Enable L3GD20 3-axis digital gyroscope library and example

    L3GD20 is tri-axis gyroscope from STMicroelectronics.
    This sensor can measure angular velocity in degree per second.

    The library provided is libupm-l3gd20.so.0.4.0.
    The example provided is l3gd20-example where it will print x,y,z axis when
    trigger buffer data is ready.

    This sensor requires calibration to be done for 2 seconds. Please place the
    sensor on level surface.

    As the sensor data is noisy, we have implemented denoise algorithm within the
    sensor library.

### DIFF
--- a/examples/c++/CMakeLists.txt
+++ b/examples/c++/CMakeLists.txt
@@ -265,6 +265,7 @@ if (BACNET_FOUND)
   add_example (e50hx)
 endif()
 add_example (vcap)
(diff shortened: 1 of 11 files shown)


laykuanloon:
Will submit another pull request

PR #396, opened 2016-04-21
Enable mmc35240

laykuanloon:
MMC35240: Enable MMC35240 3-axis magnetic sensor library and example

    MMC35240 is 3-axis magnetic sensor from MEMSIC.
    This sensor can measure magnetic fields within the full scale range of
    +-24 Gauss (G).

    The library provided is libupm-mmc35240.so.0.4.0.
    The example provided is mmc35240-example where it will print x,y,z axis when
    trigger buffer data is ready. It's also print azimuth value.

    This sensor requires calibration. Please shake the sensor in figure 8 pattern,
    mmc35240-example will print the calibrated level.

    As the sensor data is noisy, we have implemented denoise algorithm within the
    sensor library.

### DIFF
--- a/examples/c++/CMakeLists.txt
+++ b/examples/c++/CMakeLists.txt
@@ -265,6 +265,7 @@ if (BACNET_FOUND)
   add_example (e50hx)
 endif()
 add_example (vcap)
+add_example (l3gd20)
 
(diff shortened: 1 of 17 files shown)


laykuanloon:
Close this, will submit another one that disable java

PR #401, opened 2016-04-25
Mraa 1.0

arfoll:
Getting ready for small mraa API changes in 1.0.

Note alot of these are pretty nasty conversions, there's alot of i2c helper functions that aren't needed in UPM that do no error checking etc I left them as is.

### DIFF
--- a/src/adc121c021/adc121c021.cxx
+++ b/src/adc121c021/adc121c021.cxx
@@ -79,12 +79,20 @@ mraa_result_t ADC121C021::writeWord(uint8_t reg, uint16_t word)
 
 uint8_t ADC121C021::readByte(uint8_t reg)
 {
-  return mraa_i2c_read_byte_data(m_i2c, reg);
+  int val = mraa_i2c_read_byte_data(m_i2c, reg);
+  if (val != -1) {
+    return (uint8_t) val;
+  }
+  return 0;
 }
 
 uint16_t ADC121C021::readWord(uint8_t reg)
 {
-  uint16_t val = mraa_i2c_read_word_data(m_i2c, reg);
+  int x = mraa_i2c_read_word_data(m_i2c, reg);
+  if (x == -1) {
+    return 0;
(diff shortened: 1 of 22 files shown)


src/adc121c021/adc121c021.cxx
```diff
@@ -79,12 +79,20 @@ mraa_result_t ADC121C021::writeWord(uint8_t reg, uint16_t word)
 
 uint8_t ADC121C021::readByte(uint8_t reg)
 {
-  return mraa_i2c_read_byte_data(m_i2c, reg);
+  int val = mraa_i2c_read_byte_data(m_i2c, reg);
+  if (val != -1) {
+    return (uint8_t) val;
+  }
+  return 0;
```
rwaldron: This might be problematic, since a failure might look like a valid reading. For example, this chip's "Alert Status Register" and "Configuration Register" both have a default value `0`, so reading either and getting the default value has the same outcome as reading reading it but the read fails for whatever reason that would cause this condition: https://github.com/intel-iot-devkit/mraa/blob/6d8ec87e0e0561b27c1c57afcf47fbbd14aec3f3/src/i2c/i2c.c#L233-L244


`mraa`'s cpp API throws an exception: https://github.com/intel-iot-devkit/mraa/blob/c1465bd69463b34c477eea503f14716c70ed48d0/api/mraa/i2c.hpp#L140-L148 which is probably better?

jontrulson:
The MRAA rename of Spi::write_word to Spi::WriteWord needs to be fixed in UPM as well.

src/adc121c021/adc121c021.cxx
jontrulson: Agreed.  These older UPM drivers should be ported to use the C++ MRAA API at some point.

pylbert:
Merged.  We can work to get the rest of the error checking for the mraa calls in at a later date.

PR #402, opened 2016-04-25
eboled: change write_word to writeWord for MRAA 1.0 API change.

jontrulson:
Should be applied when MRAA is tagged 1.0.


pylbert:
Merged, thanks for the fix Jon.

PR #406, opened 2016-04-29
bmp280: Initial implementation

jontrulson:
The BMP280 is an absolute barometric pressure sensor especially
designed for mobile applications. The sensor module is housed in an
extremely compact 8-pin metal-lid LGA package with a footprint of only
2.0 × 2.5 mm2 and 0.95 mm package height. Its small dimensions and its
low power consumption of 2.7 μA @1Hz allow the implementation in
battery driven devices such as mobile phones, GPS modules or watches.

As the successor to the widely adopted BMP180, the BMP280 delivers
high performance in all applications that require precise pressure
measurement. The BMP280 operates at lower noise, supports new filter
modes and an SPI interface within a footprint 63% smaller than the
BMP180.

This driver supports both I2C and SPI operation.  However SPI is not
functional on Edison with the Arduino breakout board.

### DIFF
--- a/examples/c++/CMakeLists.txt
+++ b/examples/c++/CMakeLists.txt
@@ -265,6 +265,7 @@ if (BACNET_FOUND)
   add_example (e50hx)
 endif()
 add_example (vcap)
+add_example (bmp280)
 
 # These are special cases where you specify example binary, source file and module(s)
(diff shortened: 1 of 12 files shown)


jontrulson:
Will add bme280 support and resubmit when done.

PR #412, opened 2016-05-03
[ST7735] Free pointers from spi.write

drcrallen:
Docs for MRAA say ` The pointer return has to be free'd by the caller`


pylbert:
Merged.  Thanks Charles.

PR #415, opened 2016-05-04
Remove old bme280, add new combination bmp/bme280

jontrulson:
This PR has 3 commits:

The first one removes the older, somewhat broken bme280 driver.

The second adds a new combination bme280 and bmp280 driver that is fully functional, into a new bmp280 UPM library.  These drivers do not rely on the Bosch driver.

The third commit re-adds the bme280 support to the pressure, humidity, and temperature interface c++ examples that were removed in the first commit.


pylbert:
Merged.  Thanks Jon.

PR #417, opened 2016-05-04
bmp180/bmpx0x: Remove old bmp180 driver

jontrulson:
This commit removes the old bmp180 driver that was erroneously
re-added when the interface examples and headers were introduced.

It then switches the users of bmp180 to the existing bmpx8x driver,
which already supports the bmp180.


pylbert:
Merged.  Thanks Jon.

PR #423, opened 2016-05-24
bmx055, bmi055, bmc160, bma250e, bmg150, bmm150: Initial implementation

jontrulson:
This module (bmx055) implements support for the following core Bosch
chipsets:

bma250e - accelerometer, 3 variants (chip id's 0x03, 0xf9, and 0xfa)
bmm150 - magnetometer
bmg160 - gyroscope

The other 3 devices are combinations of the above:

bmx055 - accel/gyro/mag
bmc160 - accel/mag
bmi055 - accel/gyro

...for 6 devices total.

For the combination devices, all of the sub-devices appear as
individual independent devices on the I2C/SPI bus.

The combination drivers provide basic configuration and data output.
For more detailed control as well as interrupt support, you should use
the core device drivers (accel/gyro/mag) directly.

These devices support both I2C and SPI communications.  They must be
powered at 3.3vdc.


jontrulson:
merged.

Issue #280, opened 2015-10-02
JHD1313M1 not initializing correctly when Edison board powered over USB


petreeftime:
After booting the Edison board on USB power, on first run of a program requiring the JHD1313M1 LCD, the LCD does not display the first letter on the first line and the second line. On second run the LCD initializes correctly and both lines are displayed completely.

If Edison is powered using a power adapter the issue does not happen at all.

Propanu:
Yes, the RGB LCD is a 5V device and it does rely on an external power supply to work properly. I will add this issue to the documentation. When the board is powered through USB it really depends on the port, some computers/chargers will actually provide enough current for the LCD to function properly while others won't.

alext-mkrs:
I've just came across this as well. USB-powered Edison with Grove LCD, using Python. After the board power-on, running the program once results in a second row not being initialized, only the first one works in full. Second and subsequent runs of the same program initialize it properly, until the whole board is powered off (just unplugging the LCD is not enough).

As a workaround I've added double initialization to my code, which looks weird, but works. Because of this I tend to think it may be fixed just by changing timings in https://github.com/intel-iot-devkit/upm/blob/master/src/lcd/jhd1313m1.cxx - my guess is that on USB power it gets just slightly less than necessary current or voltage and because of that the controller init runs slower or something along these lines.

I'll see if I have time to play with that and propose a patch, but you may be faster.

Workaround (Python) that worked for me:
```python
grove_lcd = lcd.Jhd1313m1(0)
del self.grove_lcd
time.sleep(0.5)
grove_lcd = lcd.Jhd1313m1(0)
```

alext-mkrs:
Looks like it's not that easy as just those timings :smiley: I've checked that even increasing all timings in the constructor four-fold doesn't help it (and they're actually well above the spec limits anyway).

The voltage on the power pins of the LCD is 4.59V in my setup - barely above the spec'd 4.50V as a minimum.

I'll be looking further into this.

Propanu:
From my experience it will work OK but with washed out letters around the 4.74V that you get on a decent USB port. I don't recall if I had to initialize it twice but it's 'usable' and I got no missing letters ever since we patched the write function with a short delay. Once I plug in the external adapter into the board my reference jumps to about 4.97V and the screen is crystal clear without any issues whatsoever.

Nevertheless, the issue was added to our list of [known limitations](../blob/master/docs/knownlimitations.md)

PR #425, opened 2016-05-31
Added work around of hardware errata on Grove RGB back light LCD

mcd500:
Hello,

I have added work around of hardware errata of HD44780 LCD chip used on Grove RGB back light LCD. The HD44780 clones are widely used on many dot-matrix LCD and all has the same errata.
Without this patch, once out of five fails to initialize the HD44780 and often do not show second row or show dark background.

The datasheet of HD44780 was originally written in Japanese but this link bellow has English description of initializing the HD44780.
http://elm-chan.org/docs/lcd/hd44780_e.html

Also, the Seeed studio have already added this work around in there Arduino code.
https://github.com/Seeed-Studio/Grove_LCD_RGB_Backlight/blob/master/rgb_lcd.cpp

I really appreciate for merging my pull request.

I found this issue when I wrote the program on 96Boards at the bellow link.
https://github.com/96boards/Starter_Kit_for_96Boards/tree/master/rgb_lcd_demo

### DIFF
--- a/src/lcd/jhd1313m1.cxx
+++ b/src/lcd/jhd1313m1.cxx
@@ -46,13 +46,19 @@ Jhd1313m1::Jhd1313m1(int bus, int lcdAddress, int rgbAddress)
                                     ": I2c.address() failed");
     }
 
+    /* HD44780 requires writing three times to initialize or reset
+       according to the hardware errata on page 45 figure 23 of
+       the Hitachi HD44780 datasheet */
+    /* First try */
     usleep(50000);
-    ret = command(LCD_FUNCTIONSET | LCD_2LINE);
-
-    if (!ret) {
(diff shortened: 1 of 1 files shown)


alext-mkrs:
That sounds like a solution for #280 I was trying to find, very nice!

Looks like I was on almost the right path, just that it's not the *timing* for steps, but *number of repeats* for a certain step that works that around :smiley:

Propanu:
Thanks a lot, this is a great fix! We'll get it merged for our next UPM version.

mcd500:
It's my pleasure. I am considering adding more sensor devices to the libupm. Thank you too.

malikabhi05:
Thanks a lot!! Merged.

PR #428, opened 2016-06-10
Pull bacnetmstp update

jontrulson:
This PR updates the bacnetmstp library to support new functionality, and adds a bacnetutil class for use with BACnet drivers.


jontrulson:
merged.

PR #430, opened 2016-06-10
t8100: Initial implementation

jontrulson:
This module implements support for the Amphenol Telaire Ventostat
T8100 Ventilation Controller with BACnet interface.  It may also
support the T8200 and T8300 models, but they have not been tested.

The Telaire Ventostat T8100 reports Temperature, Humidity and CO2
concentrations.  It supports an optional relay with a settable trigger
point.  The unit this driver was tested under did not support the
optional relay.  The temperature range supported is 0-50C, humidity is
0-100% non-condensing, and CO2 range is appoximately 0-2000 PPM for
the T8100.  Other sensors in this family support wider ranges.

This PR requires that PR #428 be merged first. CI will fail on this PR.


jontrulson:
merged.

PR #431, opened 2016-06-10
tb7300: Initial implementation

jontrulson:
This module implements support for the Honeywell TB7300 Communicating
Fan Coil Thermostat.  It may also support the TB7200, though only the
TB7300 was available for development of this driver.

The TB7200 Series PI thermostats are designed for zoning applications,
and the TB7300 Series PI thermostats are designed for fan coil
control. Both Series are communicating thermostats with models
available in BACnet® MS/TP protocol and can be easily integrated into
a WEBs-AX building automation system based on the NiagaraAX® platform.

TB7200 and TB7300 Series thermostats are compatible with the Honeywell
Occupancy Sensor Cover. Thermostats equipped with an occupancy sensor
cover provide advanced active occupancy logic, which will
automatically switch occupancy levels from Occupied to Stand-By and
Unoccupied as required by local activity being present or not. This
advanced occupancy functionality provides advantageous energy savings
during occupied hours without sacrificing occupant comfort. All
thermostats can be ordered with or without a factory installed PIR
cover.

This PR requires that PR #428 be merged first. CI will fail on this PR.


jontrulson:
merged.

PR #433, opened 2016-06-18
Modif to prevent crash on the first readReg() for HTU21D

g-vidal:
Modifications to get rid of the side-effect thet occured after
commit on mraa number 1e4516d0266679a67ad8487d6d17448b76d8c211
by Brendan le Foll on 2016 Apr 22

HTU21D-F was crashing on readReg() when register was unclean
(most of the time for the first read on the i2c bus)
Changed reg type from int to uint8_t
affected initial value 0x0 to data value in resetSensor() function

### DIFF
--- a/src/htu21d/htu21d.cpp
+++ b/src/htu21d/htu21d.cpp
@@ -2,6 +2,12 @@
  * Author: William Penner <william.penner@intel.com>
  * Copyright (c) 2014 Intel Corporation.
  *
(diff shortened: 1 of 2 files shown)


Propanu:
Hi @g-vidal can you please sign-off this commit for me? Should be very easy to do with `git commit --amend -s --no-edit`. Thanks in advance!

Propanu:
Fix for this provided by 77fbf49f939e316d05942bd7752db29dd152ce60.

PR #437, opened 2016-07-04
apds9930: enable sensor before read out sensor data

laykuanloon:
Changes in library:
- enableProximity() function is to enable or disable proximity sensor
- enableIlluminance() function is to enable or disable illuminance sensor
- run clang-format

Changes in example:
- proximity and illuminance kernel IIO-based driver init state is power off,
  require enable before read out sensor data. Sleep time is needed after
  enable, the sleep time may vary on different platform.
- run clang-format

Signed-off-by: Lay, Kuan Loon <kuan.loon.lay@intel.com>


Propanu:
Merged, thanks for the updates!

PR #441, opened 2016-07-22
ublox6: Add most popular baud rates to GPS param constants

deadprogram:
Adds the most common baud rates (2400, 4800, 19200, & 38400) to the GPS param constants, since this is needed to use any GPS with a baud rate other than B9600, which previously was the only allowed value.


Propanu:
Merged, thank you for expanding these!

PR #451, opened 2016-08-12
ainBase: Abstract analog base class

pylbert:
Implemented an abstract base class for analog sensors.  This class
provides standard methods for reading from analog pins as well as
applying a scale/offset.

Updated the dfrph sensor to use the ainBase.


Propanu:
This would need some clean-up and should be split into separate commits based on what is being added, when we merge the C sources.

Also, I feel like the functionality of the analog base class, specifically scale & offset would make more sense in MRAA. @arfoll thoughts?

PR #458, opened 2016-08-25
Updates and cleanup for htu21d sensor code

whpenner:
I fixed the 16-bit read code to use what is called out in the docs (16-bit read instead of 2 8-bit ones).  There are other cleanups to the files to make it easier to use.  Code added to add the Dew Point calculation from the datasheet.  Also added a function to make it easier to read the parameters (one call instead of 3).


whpenner:
I guess you can see I'm not yet a pro at this ;-).  I had made a change some time ago in the am2315 sensor to move to the mraa_set_priority and forgot to file a pull request.

Propanu:
Merged, thanks a lot!

PR #464, opened 2016-09-23
modbus: Add version check to modules using modbus

tingleby:
Only adds the check to modules that require
MODBUS_MAX_PDU_LENGTH which was added in libmodbus
versions greater than 3.1.2

Signed-off-by: Thomas Ingleby <thomas.ingleby@intel.com>

### DIFF
--- a/src/h803x/CMakeLists.txt
+++ b/src/h803x/CMakeLists.txt
@@ -4,7 +4,7 @@ set (module_src ${libname}.cxx)
 set (module_hpp ${libname}.hpp)
 
 pkg_check_modules(MODBUS libmodbus)
-if (MODBUS_FOUND)
+if (MODBUS_FOUND AND (NOT (MODBUS_VERSION LESS 3.1.2)))
   set (reqlibname "libmodbus")
   include_directories(${MODBUS_INCLUDE_DIRS})
   upm_module_init()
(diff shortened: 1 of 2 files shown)


jontrulson:
Hi,
I think pkg_check_modules() already supports version checking, so you should be able to change the existing call to:
`pkg_check_modules(MODBUS libmodbus>=3.1.2)`
instead.

tingleby:
Yeah it does, but as its being used multiple times with the same prefix, Places where it doesn't need the version check may do something with the cache/scope and set MODBUS_FOUND to 1 regardless to version.
```
pkg_check_modules(H803XLIBS libmodbus>=3.1.2)
```
Would also work pretty nicely but I didnt want to mess around with things too much.

jontrulson:
I see your point, but that's not actually a problem.  All of the current MODBUS drivers were developed with 3.1.2, so in reality, I should have made this a requirement (via pkg_check_modules) from the very beginning.  I even went to the trouble to specify the version requirement in the documentation :)  My bad.

So, I'd prefer we just change the pkg_check_modules() to require >=3.1.2 in all of the MODBUS drivers (there are only 3 - h803x, hwxpxx, and t3311).

Propanu:
There are inconsistencies between pkg_check_modules and find_package. In order to work around I'm tempted to pull all 3rd party library requirements to the top-level CMake and in the sensor CMakes themselves only to add the LIBX_FOUND check.

jontrulson:
No opposition from me...

tingleby:
@Propanu That makes more sense for packaging too.

tingleby:
Top Level is better :D

PR #467, opened 2016-10-04
python_examples: Reformatted, self-checking, executable

pylbert:
* Moved body of each python example to main.  This allows for basic
      load module testing for CI
    * General cleanup of python modules (crlf/tabs/prints/etc)
    * Chmod'ed to 755 to allow running examples without specifying the
      python interpreter
    * Added ctest for loading python2/3 modules
    * Added jniclasscode pragma for java swig interface files.
    * Updated check_examplenames.py module to check all languages vs. a
      cxx example name
    * Added tests for checking python module and test loading
    * Added 'make test' to travis-ci run (run ctests)
    * Print a more meaningful message when not building cxx docs into
      python modules
    * Updated check_clean.py to only check java wrapper files.
    * ENABLED ctests for UPM


examples/python/using_carrays.py
```diff
@@ -24,8 +24,11 @@
```
jontrulson: Hmm, is it even worth keeping this example?

src/lp8860/javaupm_lp8860.i
```diff
@@ -6,3 +6,14 @@
 %}
 
 %include "lp8860.hpp"
+
```
jontrulson: Ideally these fixes would be a separate PR, but no biggie.

tests/check_examplenames.py
```diff
@@ -0,0 +1,75 @@
...
+# Example name map
+sampleMappingFile = '../doxy/samples.mapping.txt'
+
+# Example directories
+sampledir = {}
+sampledir['cxx'] = '../examples/c++/'
+sampledir['java'] = '../examples/java/'
+sampledir['js'] = '../examples/javascript/'
+sampledir['py'] = '../examples/python/'
+
```
jontrulson: Should this handle C examples as well?

tests/check_pythonload.py
```diff
@@ -0,0 +1,68 @@
...
+             # Requires mraa python module
+             'curieimu',
+             # Requires modbus
+             'h803x', 'hwxpxx', 't3311',
+             # Requires bacnet
+             'e50hx', 'tb7300', 't8100',
+             # Requires PIL
+             'make_oled_pic'
+             ]
+
```
jontrulson: Would it be better to have these in a datafile rather than hardcoded into a script?

doxy/samples.mapping.txt
```diff
@@ -1,73 +1,72 @@
...
-stepmotor.cxx	StepMotorSample.java
-tm1637.cxx	TM1637Sample.java
-tp401.cxx	TP401Sample.java
-tsl2561.cxx	TSL2561Sample.java
-ttp223.cxx	TTP223Sample.java
-ublox6.cxx	Ublox6Sample.java
-uln200xa.cxx	ULN200XASample.java
-wt5001.cxx	WT5001Sample.java
-yg1006.cxx	YG1006Sample.java
+# Mapping examples across provided languages
```
jontrulson: Interesting, I was not aware of this file.  No doubt it is incomplete.  What about C examples?  How are cases handled where a driver has more than one example (urm37 vs. urm37-uart, for example).

examples/python/aeotecdsb09104.py
```diff
@@ -24,65 +24,50 @@
...
-
-        print "Energy, Channel 3:",
-        print sensor.getEnergyC3(),
-        print "kWh"
-
-        print "Battery Level:",
-        print sensor.getBatteryLevel(),
-        print "%"
-        print
+        print "Battery Level: %0.03f\n" % sensor.getBatteryLevel()
```
pylbert: Fixed...

examples/python/using_carrays.py
pylbert: Not sure what this was ever used for.  If it's just there to show how to use a swig carray type then I agree.

tests/check_pythonload.py
pylbert: I would like to keep from adding additional files for testing.  One thought is to install the tests and examples according to the cmake options and run from the install directory.  In this way, if tpv libraries were not available during the cmake generation phase, then the installed examples would reflect this (and there would be no need for a blacklist in this case).

examples/python/using_carrays.py
Propanu: OK to remove, we have plenty of examples for this within the drivers that actually use it.

pylbert: Removed...

PR #468, opened 2016-10-12
python: Added upm directory for python modules

pylbert:
* Grouped UPM python modules into upm directory, for example:
      /usr/local/lib/python2.7/dist-packages/upm
    * Updated UPM example import statements
    * Removed unused RPATH statements from UPM src CMakeLists.txt,
      currently build collateral contains an explicit RPATH which
      is stripped from the install collateral.
    * Converted python examples to work on both python2 AND python3
    * Added ctest for loading examples w/python3
    * Removed returns from swig macros

### DIFF
--- a/examples/python/aeotecdw2e.py
+++ b/examples/python/aeotecdw2e.py
@@ -21,13 +21,14 @@
 # OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
 # WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
+from __future__ import print_function
 import time, sys, signal, atexit
-import pyupm_ozw as sensorObj
+from upm import pyupm_ozw as sensorObj
 
 def main():
     # This function lets you run code on exit
     def exitHandler():
-        print "Exiting"
+        print("Exiting")
         sys.exit(0)
 
     # Register exit handlers
@@ -36,7 +37,7 @@ def main():
     defaultDev = "/dev/ttyACM0"
     if (len(sys.argv) > 1):
         defaultDev = sys.argv[1]
-    print "Using device", defaultDev
+    print("Using device", defaultDev)
 
     # Instantiate an Aeotec Door/Window 2nd Edition sensor instance, on
     # device node 10.  You will almost certainly need to change this to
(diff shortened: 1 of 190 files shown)


examples/python/wt5001.py
```diff
@@ -53,7 +54,7 @@ def printUsage(progname):
     elif cmd == 2:
         myMP3Player.pause()
     elif cmd == 3:
-        myMP3Player.next()
+        next(myMP3Player)
```
jontrulson: I don't think this is correct :) next() tells the player to skip to the next track.

src/CMakeLists.txt
```diff
@@ -38,35 +38,25 @@ macro (upm_target_link_libraries target_name)
 
     # Call the swig link library method for each library
     if (BUILDSWIGNODE)
-      set_target_properties(${SWIG_MODULE_jsupm_${libname}_REAL_NAME} PROPERTIES SKIP_BUILD_RPATH TRUE)
       swig_link_libraries (jsupm_${libname} ${_library})
```
jontrulson: For these RPATH changes - you might ask Tudor about them as I think he added them originally.  I believe they were to work around issues with the eclipse IDE(s), but I don't know the specifics.

Propanu: This is only for node modules thus Eclipse wouldn't be impacted. However, using SKIP_BUILD_RPATH TRUE will likely remove the path for both build & install. Ideally we want the path for the build (in case tests need to run) and only stripped when the module is installed.

examples/python/aeotecdw2e.py
```diff
@@ -49,32 +50,32 @@ def exitHandler():
...
+            print("%")
 
-            print
+            print()
         else:
-            print "Device has not yet responded to probe."
-            print "Try waking it, or wait until it wakes itself if ",
-            print "configured to do so."
+            print("Device has not yet responded to probe.")
+            print("Try waking it, or wait until it wakes itself if ", end=' ')
```
pylbert: Good catch, but for some weird reason, this is how this python print statement was written.

examples/python/e50hx.py
```diff
@@ -69,30 +70,30 @@ def exitHandler():
...
 
     # update and print available values every second
     while (1):
-        print "System Voltage:",
-        print sensor.getAnalogValue(sensorObj.E50HX.AV_System_Voltage),
-        print " ",
-        print sensor.getAnalogValueUnits(sensorObj.E50HX.AV_System_Voltage)
+        print("System Voltage:", end=' ')
+        print(sensor.getAnalogValue(sensorObj.E50HX.AV_System_Voltage), end=' ')
+        print(" ", end=' ')
```
pylbert: Same here

examples/python/e50hx.py
```diff
@@ -69,30 +70,30 @@ def exitHandler():
...
+        print(sensor.getAnalogValue(sensorObj.E50HX.AV_System_Type))
 
-        print "Energy Consumption:",
-        print sensor.getAnalogInput(sensorObj.E50HX.AI_Energy),
-        print " ",
-        print sensor.getAnalogInputUnits(sensorObj.E50HX.AI_Energy)
+        print("Energy Consumption:", end=' ')
+        print(sensor.getAnalogInput(sensorObj.E50HX.AI_Energy), end=' ')
+        print(" ", end=' ')
+        print(sensor.getAnalogInputUnits(sensorObj.E50HX.AI_Energy))
```
pylbert: And here

examples/python/wt5001.py
pylbert: Fixed

examples/python/e50hx.py
jontrulson: ?  What am I missing... Original:

`        print "System Voltage:",`
`        print sensor.getAnalogValue(sensorObj.E50HX.AV_System_Voltage),`
`        print " ",`
`        print sensor.getAnalogValueUnits(sensorObj.E50HX.AV_System_Voltage)`

All on one line, separated by spaces.  New code adds 3 spaces between value and units...

pylbert: Ugh, the snippet I'm pasting in has spaces between the 10.0 values, not sure why it doesn't print like that in github

jontrulson: Yep, I got it.  Probably need a `code` block.

examples/python/aeotecdw2e.py
jontrulson: Oh, and the reason it was written that way originally was to avoid line wrapping (this is/was a requirement).

pylbert:
Consider using some of the other methods for string concat/wrapping in
python.  Example:

>>> print 'this is a long string ' \
...     'that continues on another line'
this is a long string that continues on another line

or

>>> print ('this is a long string '
...     'that continues on another line')
this is a long string that continues on another line

As well as os.linesep or '\n's instead of added empty print statements
(which is nicer to read IMO).

jontrulson:
Agreed, will do.

Propanu:
I'd say it's pretty safe to merge this as we don't have an IDE integrated with Python, although please do a check first on the other `intel-iot-devkit` samples repositories that might be using UPM Python modules.

PR #471, opened 2016-10-20
C11 changes

jontrulson:
These commits unconditionally enable C11 standards compliance, and fix up 2 errors uncovered.

### DIFF
--- a/src/rotary/rotary_fti.c
+++ b/src/rotary/rotary_fti.c
@@ -26,6 +26,7 @@
 
 #include "rotary.h"
 #include "upm_fti.h"
+#include "upm_math.h"
 
 /** 
  * This file implements the Function Table Interface (FTI) for this sensor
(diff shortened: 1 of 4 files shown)


src/rotary/rotary_fti.c
```diff
@@ -27,6 +27,11 @@
 #include "rotary.h"
 #include "upm_fti.h"
 
+// M_PI is not a C standard (C11)
+#ifndef M_PI
+#define M_PI           3.14159265358979323846
+#endif
+
```
Propanu: This would go into `upm_math.h` as it's used by multiple sensors, exactly why I added the file for 😉.

jontrulson: Will fix, thanks!

jontrulson:
Merged, thanks.

PR #473, opened 2016-10-25
Add support to integrate upm drivers in Android

spitfire88:
Fix compilation issues.

Signed-off-by: Sanrio Alvares sanrio.alvares@intel.com

### DIFF
--- a/Android.mk
+++ b/Android.mk
@@ -1,0 +1,61 @@
+#
+#  Author: Sanrio Alvares <sanrio.alvares@intel.com>
+#  Copyright (c) 2016 Intel Corporation.
+#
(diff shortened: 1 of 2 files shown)


spitfire88:
With this commit, UPM should be able to run in an Android environment. Should we maintain a different branch for this? The current CI tests will fail unless tested in an Android Environment.

Propanu:
Will this build a singleton for Android? I don't see any reasons for Travis breaking, as the makefiles won't be used by the Travis build.

However, it would be very difficult to maintain this list manually. Thus we should look into having the `upm_src_files` list and consequently, the makefile, generated by cmake instead.

malikabhi05:
UPM is now supported on Android Things.

PR #474, opened 2016-10-26
Incorrect Function Name

SuyashLakhotia:
`groveSensor.Button()` --> `groveSensor.GroveButton()`

References for Change:
- http://iotdk.intel.com/docs/master/upm/node/classes/grovebutton.html
- https://software.intel.com/en-us/iot/hardware/sensors/grove-button

### DIFF
--- a/examples/javascript/button.js
+++ b/examples/javascript/button.js
@@ -26,7 +26,7 @@
 var groveSensor = require('jsupm_grove');
 
 // Create the button object using GPIO pin 0
-var button = new groveSensor.Button(0);
+var button = new groveSensor.GroveButton(0);
 
 // Read the input and print, waiting one second between readings
 function readButtonValue() {


examples/javascript/button.js
```diff
@@ -26,7 +26,7 @@
 var groveSensor = require('jsupm_grove');
```
jontrulson: Hi, I think this is the real bug - instead of loading jsupm_grove, this example should be loading jsupm_button.

Can you give that a try and see if that works with the original Button(0) below?

SuyashLakhotia: Hey, that might work (will try once I get my hands on the Edison again and update here) but requiring a separate node module for every type of sensor (especially when they're all part of the same Grove kit) seems redundant to me. Let me know what you think! :)

SuyashLakhotia: Additionally, the [Intel website](https://software.intel.com/en-us/iot/hardware/sensors/grove-button) uses `jsupm_grove` too.

Propanu: @jontrulson you're right, @SuyashLakhotia any chance you can open this as an issue instead please?

SuyashLakhotia: Ah, I just realised that there are separate examples for [`button.js`](https://github.com/intel-iot-devkit/upm/blob/master/examples/javascript/button.js) and [`grovebutton.js`](https://github.com/intel-iot-devkit/upm/blob/master/examples/javascript/grovebutton.js). I see what you mean now.

@Propanu Sure, I'll open an issue.

SuyashLakhotia: @jontrulson @Propanu Opened at #480.

PR #479, opened 2016-11-02
java: Removed mraajava dependency, replaced with swig implementation

pylbert:
* Added java_isr.i interface file which defines a C++ proxy
      class for C++ <-> java callbacks
    * Proxy class is called UpmCallback and java can inherit/override
      run method via directors
    * Removed ifdefs for JAVACALLBACK from source
    * Updated java examples to use UpmCallback (MMA7660Sample.java and
      Button_intrSample.java)


pylbert:
The intent of this PR is to POC an alternate method for JAVA <-> C callbacks for UPM sensor libraries.

I would like to update all UPM libraries which rely on mraajava to instead use this callback method and remove our dependency on mraajava.

Am I missing any functionality here?

If we go this route and all looks good, I can make the necessary edits to UPM drivers.

whbruce:
I like this idea. Here are some ideas for further improvements.
- Pass through GPIO number so callback receives it
- Is is possible to create a common (module independent) callback base class (e.g upm.GpioISR)
- Rename Java "callback" method to something more meaningful like onGpioISR(int pin)

Propanu:
Here's the fine print:
Mraajava.so was linked to all UPM Java Native libraries because, some time ago, there was a common class defined in MRAA (called IsrCallback) which was being used as a base class for all callbacks for all sensors, since there was no common UPM package.

Longer story:
This IsrCallback class in MRAA was later replaced by changing all UPM & MRAA Java interfaces/code to take directly java.lang.Runnable objects as callbacks. This resulted in a pretty big performance improvement when handling callbacks in Java.

IMHO this will take us right back to where we were with the IsrCallback class, albeit without the MraaJava dependency. We're talking 2.6x ISR throughput increase with runnables vs. proxy class so needs to be double checked.

jontrulson:
I like this too, as it removes java-specific ifdef checks in the drivers.  However, if @Propanu is correct regarding the performance, we may have to keep it as it is... 2x drop is too big.

pylbert:
Closing - I see the performance delta.  Maybe this can be profiled at a later point by someone who cares about the JAVA implementation.

PR #490, opened 2016-11-21
cmake: Move find package early to avoid unsetting cross compile flags…

arfoll:
… in configure stage

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,7 +89,7 @@ function (upm_add_compile_flags compiler)
   endforeach (flag ${ARGN})
 
   # Set the variable in the parent scope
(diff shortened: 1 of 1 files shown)


arfoll:
With the following fix CFLAGS are appended so this stops this issue.

pylbert:
This should have been fixed here: https://github.com/intel-iot-devkit/upm/commit/cdc33ceb9fd592e82c1eb940dbac5cb0fbfeded2#diff-af3b638bc2a3e6c650974192a53c7291R89 

Let me know if this is not the case.

Propanu:
Is this confirmed fixed?

pylbert:
Fixed in cdc33ceb9fd592e82c1eb940dbac5cb0fbfeded2

PR #491, opened 2016-11-21
Update README.md

fromgilbert:
add ppa install instructions


Propanu:
@fromgilbert thanks a lot for this, I will create a separate page with installation instructions for the different OS'es that we support now. I think this was long due so I appreciate you taking the first steps.

Propanu:
Added, thanks again!

PR #500, opened 2016-12-05
LICENSE: Update license to be clearer and add 2016 copyright

arfoll:
MIT licenses should start with the following line to identify it as an MIT
license.


Propanu:
Merged, thanks! Might as well update it to 2017 now 😄.

PR #519, opened 2017-02-07
cmake: Explicit dependency for mraa and interfaces

pylbert:
A handful of modules do not require mraa.  Captured this in
    src/CMakeLists.txt - only add mraa dependency for targets which
    use:

        upm_module_init(mraa ... )
            or
        upm_mixed_module_init(... REQUIRES mraa)

    All sensors which use UPM interfaces (src/interfaces) now
    explicitly add the interfaces target:

        upm_module_init(interfaces ... )
            or
        upm_mixed_module_init(... REQUIRES interfaces)


src/CMakeLists.txt
```diff
@@ -302,7 +302,7 @@ macro(upm_swig_java)
     # See issue #518
     # TODO: Fix this somehow
     swig_add_module (javaupm_${libname} java javaupm_${libname}.i ${module_src})
-    swig_link_libraries (javaupm_${libname} ${MRAAJAVA_LIBRARIES} ${MRAA_LIBRARIES} ${JAVA_LIBRARIES} ${libname} interfaces)
+    swig_link_libraries (javaupm_${libname} ${MRAAJAVA_LIBRARY} ${JAVA_LIBRARIES} ${libname} interfaces)
     target_include_directories ( ${SWIG_MODULE_javaupm_${libname}_REAL_NAME}
```
jontrulson: Hmm... I'm wondering if an MRAA dependency should just automatically include the MRAAJAVA dependency, as I think it's required if using Java and MRAA, but not required at all if your driver (and therefore Java example) is not using MRAA...?

pylbert:
I was leaving this as a TODO since I don't have a good solution.  I would
like to have a cmake way of recursively finding all dependencies for
${libname}, then, if mraa is there, also link to mraajava.

pylbert:
@jontrulson, ppd42ns sets a dependency on the c target.  I think this is a good way to do it since the CXX target depends on the C target.  Another way would be to add 'm' to requires which would add the dependency directly to both.  We could possibly add a CPP_REQUIRES and a C_REQUIRES.

PR #521, opened 2017-02-08
SWIG_JAVA: C++ Interfaces to Java interfaces

sasmita:
This commit translates C++ interfaces to Java interfaces, previously C++ Interfaces implemented java classes.

    * Added java swig interface files for all C++ interfaces to simplify swig
    javaupm_iADC.i
    javaupm_iCO2Sensor.i
    javaupm_iHumiditySensor.i
    javaupm_iLightController.i
    javaupm_iLightSensor.i
    javaupm_iModuleStatus.i
    javaupm_interfaces.i
    javaupm_iPressureSensor.i
    javaupm_iTemperatureSensor.i

    <example Usage>
        %include"../interfaces/javaupm_iADC.i"

    * Modified swig interface files for few sensors that implements interfaces
    ads1x15
    bmp280
    bmpx8x
    ds1808lc
    hlg150h
    lp8860
    max44009
    ms5611
    si1132
    si7005
    t6713

    * Removed few methods that were mentioned Protected and made them public, so that menthods can be overridden

    * Made IModuleStatus virtual to avoid ambiguity in multiple inheritance
    For example
    class A {};
    class B : public A {};
    class C : public A {};
    class D : public B, public C {};

    This can be solved as
    class A {};
    class B : virtual public A {};
    class C : virtual public A {};
    class D : public B, public C {};

Signed-off-by: sisinty sasmita patra <sisinty.s.patra@intel.com>


whbruce:
I'm pleased that someone is looking into making Java interfaces work, I could never figure it out.
To prove it works, I'd like to see a Java example similar in nature to [light-sensor.cxx](https://github.com/intel-iot-devkit/upm/blob/master/examples/c%2B%2B/light-sensor.cxx)

examples/java/BME280_Example.java
```diff
@@ -32,6 +32,9 @@ public static void main(String[] args) throws InterruptedException
 
         // Instantiate a BME280 instance using default i2c bus and address
         BME280 sensor = new BME280();
+        ITemperatureSensor t_sensor = sensor;
+        IHumiditySensor h_sensor = sensor;
+        IPressureSensor p_sensor = sensor;
```
sasmita: @whbruce, Here is a java example that uses interfaces.

stefan-andritoiu:
Also because of this, interfaces can not be used to implement collections of different classes (e.g. a list of different temperature sensors that all have a getTemperature() method), and only serve as an abstraction to one sensor class (which can be useful, but not the only goal of a Java interface)

PR #529, opened 2017-03-10
java: solved unclean bindings

stefan-andritoiu:
Signed-off-by: Bogdan Ichim <bogdan.ichim@rinftech.com>
Signed-off-by: Stefan Andritoiu <stefan.andritoiu@gmail.com>

### DIFF
--- a/src/htu21d/javaupm_htu21d.i
+++ b/src/htu21d/javaupm_htu21d.i
@@ -1,5 +1,13 @@
 %module javaupm_htu21d
 %include "../upm.i"
+%include "typemaps.i"
+
+%apply float *OUTPUT {float *fHum};
+%apply float *OUTPUT {float *fHumTemp};
+%apply float *OUTPUT {float *fDewPt};
+%inline %{
+    extern int getHumidityData(float* fHum, float* fHumTemp, float* fDewPt);
+%}
 
 %{
     #include "htu21d.hpp"
--- a/src/ozw/javaupm_ozw.i
+++ b/src/ozw/javaupm_ozw.i
@@ -5,6 +5,12 @@
 %include "arrays_java.i";
 %include "../java_buffer.i"
 
+%apply unsigned char *OUTPUT { uint8_t *val };
+
+%inline %{
+    extern void setValueAsBytes(int nodeId, int index, uint8_t *val, uint8_t len);
+%}
+
 %include "ozw.hpp"
 %{
     #include "ozw.hpp"


stefan-andritoiu:
If the native code has been split into C and C++, I see no reason to export the low-level C code to Java, when the C++ classes cover all possible functionality in a cleaner non-pointer way.
Therefore I propose eliminating from the SWIG interfaces the *%include "sensor.h"* part if *%include "sensor.hpp"* is present

jontrulson:
I agree with the principle, but not necessarily with the solution... Some of the earlier C ports did this - defined enumerations and #defines in the .h file (just like they originally did in the .hpp file), so they would be available to the end-user in the SWIG languages.  Of course, then these .h files were added to the SWIG interface files so they would be available.

I agree, this is incorrect - but I think it should be addressed at the driver level, rather than duplicate "some" of the enums in the interface file.

Newer C ports (at least the ones I am doing) split out these components (#defines, enums) in a separate driver_regs.h, or driver_defs.h file, which can be "swig-a-fied" safely.  This allows the constants to be made available to the end user, without also including C function definitions in the SWIG output.

I would suggest those drivers that are problematic in this respect be fixed (by splitting constant defines for which the user may want/need access to into a separate _defs.h or whatever is appropriate for the situation.  

See ds18b20_defs.h (ds18b20), bno055_regs.h (bno055), and others for examples.  In practice, this seems to work well, even in Java.

jontrulson:
I will look at this in the next few days, and submit a new PR that should hopefully resolve most of these issues.  I'd appreciate if you could check it out/test when ready.

stefan-andritoiu:
I was not aware of this plan but I totally agree that separating enums and defs in their own file is a cleaner and better solution.
This way SWIG can include them without having to include the headers of the C files and automatically the C function definitions, and also don't need to be added by hand in each interface file.

jontrulson:
@stefan-andritoiu In PR #532, I have addressed at least the header files issues.  I did not incorporate the changes you made to _javaupm_htu21d.i_ or to _javaupm_ozw.i_ as I think those are some other issue - I focused only on the SWIG .h issues.

The lcm1602 and jhd1313m1 have already been fixed in recent commits, so I did not touch those either.  Let me know if they are still a problem.

Thanks!

stefan-andritoiu:
Looks good, Jon. I think we should make this the standard way for enums and defs that should be available in all languages.
Changed this PR to only include the fixes for htu21d and ozw. PR should hove no conflict with the main upm or your PR #532

jontrulson:
> I think we should make this the standard way for enums and defs that should be available in all languages.

Agreed!  Thanks!

PR #531, opened 2017-03-16
Added bus autodetect to Intel Aero sensor C++ samples

whbruce:
bmi160, bmm150 and ms5611 C++ samples now autodetect i2c and spi busses.

### DIFF
--- a/examples/c++/bmi160.cxx
+++ b/examples/c++/bmi160.cxx
@@ -23,6 +23,7 @@
  */
 
 #include <unistd.h>
+#include <string>
 #include <iostream>
 #include <signal.h>
 #include "bmi160.hpp"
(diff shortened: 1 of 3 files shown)


examples/c++/bmi160.cxx
```diff
@@ -41,10 +42,30 @@ void sig_handler(int signo)
...
+  }
+  catch (std::exception& e)
+  {
+    std::cout << "Failed to connect via I2C, trying default SPI bus" << std::endl;
+  }
+
+  try
+  {
+    if (sensor == NULL)
+      sensor = new upm::BMI160(0, -1);
```
jontrulson: This is unlikely to work without specifying the correct CS pin.  The default is -1 (as 3rd arg), which is suitable for a device in which the OS/kernel handles CS for you (like Edison on pin 10).  But on other platforms, you would need to actually specify whatever pin you connected CS to, which can be anything.

examples/c++/bmm150.cxx
```diff
@@ -44,8 +44,25 @@ int main(int argc, char **argv)
...
+  {
+    try
+    {
+      cout << "Trying i2c address " << hex << showbase << BMM150_DEFAULT_ADDR + i << endl;
+      sensor = new upm::BMM150(BMM150_I2C_BUS, BMM150_DEFAULT_ADDR + i);
+    }
+    catch (std::exception& e)
+    {
+    }
+
```
jontrulson: This logic seems incorrect...  Say the device supports 0x1, 0x2, 0x3, and 0x4 as possible I2C addresses, and the device is actually configured for 0x02.  What happens in the above loop when you succeed?  Nothing, as it will continue trying the other addresses and eventually fail (and leak the memoy of the instance that did succeed).  You probably need a "break" in there.

In general, I am opposed to this kind of magic "guess the valid resources" logic - it's bug prone and no real guarantee that the device is actually at a specified location.  Some devices can verify if the real device is present (via a chip id or something similar), and not some other device that is using the same I2C address, and some (many, most?) cannot.  There is no implied guarantee here.

src/ms5611/ms5611.cxx
```diff
@@ -54,7 +54,7 @@ using namespace upm;
 MS5611::MS5611(int i2cBus, int address)
 {
     status = mraa::ERROR_INVALID_RESOURCE;
-    i2c = new mraa::I2c(2);
+    i2c = new mraa::I2c(i2cBus);
     this->address = address;
```
jontrulson: This definitely needs to be fixed, but perhaps in a separate PR?

pylbert: Fixed in f92c0c120e3c0ae0852ba2eebf198a7e3e6090df.

whbruce:
Not needed.

PR #539, opened 2017-04-04
Transitive examples

pylbert:
I would like to propose this change for our C/C++ examples.  This allows the examples to use our sensor libraries in a transitive way (without the 300+ -I include directories for each target compile).

Main items to review: 
    Added src/CMakeLists.txt
    Changes to src/c++/CMakeLists.txt
    Changes to src/c/CMakeLists.txt
    Renames of examples in src/c++/CMakeLists.txt
    Renames of examples in src/c/CMakeLists.txt


pylbert:
Updated the names for bma25e, bmg160, and bmm150.

I have validated/updated all the '@snippet' lines but have some questions.

Should all the ozw headers have @snippets to their corresponding examples?
I only see one in ozwdump.hpp:

https://github.com/intel-iot-devkit/upm/blob/master/src/ozw/ozwdump.hpp#L44

Example:

https://github.com/intel-iot-devkit/upm/blob/master/src/ozw/aeotecdsb09104.hpp#L39
https://github.com/intel-iot-devkit/upm/blob/master/examples/c%2B%2B/aeotecdsb09104.cxx

Also, I see examples with the ! [Interesting] tag, but the source headers
do not reference it (which I assume reads in the entire file).  See
https://github.com/intel-iot-devkit/upm/blob/master/src/ozw/ozwdump.hpp#L44.
Is this on purpose, or should I add the ! [Interesting] in these cases?

jontrulson:
> Should all the ozw headers have @snippets to their corresponding examples?

Yes.  If they are missing, it was an oversight.

> Also, I see examples with the ! [Interesting] tag, but the source headers do not reference it (which I assume reads in the entire file).

Same here.

PR #546, opened 2017-04-18
Android things

pylbert:
This PR allows building against the Android NDK (14b) platform-24 using libc++.


include/upm_fti.h
```diff
@@ -30,6 +30,9 @@
 extern "C" {
 #endif
 
+/* Get filename w/o path */
+#define __FILENAME__ (strrchr(__FILE__, '/') ? strrchr(__FILE__, '/') + 1 : __FILE__)
```
jontrulson: Should we also include string.h in this header file since we are using strrchr()?

src/htu21d/htu21d.cpp
```diff
@@ -154,12 +154,16 @@ HTU21D::getDewPoint(int bSampleData)
...
+    float fB  = 1762.39;
+    float fC  = 235.66;
+    float fT  = getTemperature(false);
+    float fRH = getHumidity(false);
+#ifdef __ANDROID__
+    float fPP = pow(10, fA - (fB / (fT + fC)));
+#else
     float fPP = exp10(fA - (fB / (fT + fC)));
+#endif
     float fDP = -(fB / (log10(fRH * fPP / 100) - fA) + fC);
```
jontrulson: I wonder if we should just stick to pow() (or perhaps, more appropriately, powf()) here and get rid of exp10()?  exp10 seems to be a gnu extension only.  Then the ifdef would not be necessary?

src/mpu9150/mpu60x0.hpp
```diff
@@ -802,19 +802,20 @@ namespace upm {
...
+//     */
+//    float *getAccelerometer();
+//
+//    /**
+//     * get the gyroscope values
+//     *
+//     * @return Array containing X, Y, Z gyroscope values
+//     */
+//    float *getGyroscope();
 #endif
```
jontrulson: Quite an oversight on my part.  We should put that on the list of things to do: fix the "Java Problem" once and for all like newer drivers do (use vectors, and have all java-specific method implementations present in the .hpp file only)

include/upm_fti.h
pylbert: Done.

src/htu21d/htu21d.cpp
pylbert: Done.

src/mpu9150/mpu60x0.hpp
pylbert: Agreed. Removed.

PR #550, opened 2017-05-16
java: modified functions that return SWIG generated vector types to return AbstractList

stefan-andritoiu:
Changes:
1. Removed unused wrapper classes for std::vector
Including upm_vectortypes.i creates all vector types (intVector, floatVector, etc) in final package, some of which are not needed and add to the footprint. Selectively added only those that are used by the package.

2. Converted SWIG-generated wrappers for std::vector to inherit AbstractList<>. This way they can be cast to it.

3. Modified functions in UPM that return vectortypes to return AbstractList.
The actual object returned is still an intVector (or floatVector, etc) but it is cast to an AbstractList<Int> (AbstractList<Float>). 

This solves the semantic ambiguity problem of having floatVector in two separate packages and needing to specify between them.

This way the user never has to work with (or even know about) the SWIG-generated wrappers for std::vector

The get/set methods that were implemented for the typeVectors work as they should on AbstractList, only difference is that now they overwrite the set/get methods of AbstractList.  

4. Modified the examples to use AbstractList instead on typeVectors


src/bma250e/javaupm_bma250e.i
```diff
@@ -1,10 +1,25 @@
...
+%typemap(javaout) upm::BMA250E {
+    return new $&javaclassname($jnicall, true);
+}
+%typemap(javaout) std::vector<float> {
+    return (AbstractList<Float>)(new $&javaclassname($jnicall, true));
+}
+%typemap(jstype) std::vector<float> "AbstractList<Float>"
+
+%template(floatVector) std::vector<float>;
+
```
jontrulson: Hi, is this stanze only needed for the Float types in Java?  I notice you did not add these for Int's and uint8_t's (chars/bytes).

This stanza also appears to be duplicated in the other java interface files dealing with Floats.  Could we put this into a separate .i file that we just include when needed, thereby keeping the driver .i's as clean as possible?

src/bno055/javaupm_bno055.i
```diff
@@ -13,6 +14,10 @@
 %ignore getLinearAcceleration(float *, float *, float *);
 %ignore getGravityVectors(float *, float *, float *);
 
+%template(floatVector) std::vector<float>;
+%template(byteVector) std::vector<uint8_t>;
+%template(intVector) std::vector<int>;
+
```
jontrulson: This one also uses floats, but the stanza mentioned above is not present... Should it be?

src/upm_stdvector.i
```diff
@@ -0,0 +1,61 @@
+%include "autobox.i"
```
jontrulson: This file appears to be almost a copy of the java/std_vector.i interface file.  This is to support the use of AbstractList?  But it is only needed for Java, so should the file be called something else, like upm_java_vector.i ?

src/bma250e/javaupm_bma250e.i
stefan-andritoiu: Hi, we only added them for Float types as that sensor only has functions returning std::vector\<float\>. We only added what was needed (not that it would make a difference if the other would be added)

As for adding these in a separate file and including them wherever needed might prove problematic for sensors that have std::vector\<\> as both input and output (see * explanation for javaupm_bno055.i).


But we could add these in a separate file and include them wherever needed (to keep those files cleaner), and treat special sensors (bno055) in their own \.i file.

src/bno055/javaupm_bno055.i
stefan-andritoiu: Yes it should, we will update this and the newly added sensors also.
This sensor only has the modifications to remove the unnecessary typeVectors classes that it didn't use but were still added to the package.

\*The problem with this class is that it has a function that takes _std::vector<uint8_t>_ as input, and this is not as straight forward to solve. If we force it to take an AbstractList<Byte> and then down cast it to a byteVector it will crash if the user gives it an AbstractList<Byte> that is not actually a byteVector in disguised.
A solution for this would be to have SWIG create a new typeVector and copy the content of the AbstractList<Byte> into it before passing the typeVector down. This way the function would take an AbstractList but pass down a typeVector for further processing.


We will add the AbstractList wrapper for all **outputs** of functions, but until we resolve this, the inputs will remain as typeVectors.

src/upm_stdvector.i
stefan-andritoiu: Yes we should rename this to either upm_javavector.i / upm_javastdvector.i / upm_stdjavavector.i
Which sounds more appropriate?

src/bno055/javaupm_bno055.i
jontrulson: ok

src/bma250e/javaupm_bma250e.i
jontrulson: That sounds good - I just want to make it as "easy" as possible going forward to support this properly in new Java examples.

src/upm_stdvector.i
jontrulson: How about upm_javastdvector.i?  Thanks for the  comments - as you can tell I'm not a experienced Java developer :)

stefan-andritoiu: Renamed it to upm_javastdvector.i

Propanu:
Merged, thanks!

Issue #547, opened 2017-04-26
Repeated I2C_SLAVE ioctls before every RD/WR fail on Android Things


spitfire88:
According to the I2C spec, if you are using the SMBus method to RD/WR you need do the I2C_SLAVE ioctl only once and then proceed with multiple read/write calls.

See here: https://www.kernel.org/doc/Documentation/i2c/dev-interface 
"_When you have opened the device, you must specify with what device
address you want to communicate:

  int addr = 0x40; /* The I2C address */

  if (ioctl(file, I2C_SLAVE, addr) < 0) {
    /* ERROR HANDLING; you can check errno to see what went wrong */
    exit(1);
  }

Well, you are all set up now. You can now use SMBus commands or plain
I2C to communicate with your device._"

Android Things and MRAA both use the SMBus method to RD/WR. But some drivers like LSM303 do not follow this spec and repeatedly set the address before every RD/WR. In fact there are 6 of them: ds1307, itg3200, lsm303dlh, pn532 and adxl345.

Why is the address set repeatedly in these drivers and not the other I2C based ones? Is it due to the “repeated start condition” in I2C? It would explain why the address is set before every RD/WR. This condition doesn't apply to SMBus RD/WR method according to the spec. It is needed in case ioctl I2C_RDWR is used for reading and writing the slave device. None of the said drivers use this method  (although it is supported in MRAA).

Can we get rid of the repeated calls?
Here is an example: 
1. https://github.com/intel-iot-devkit/upm/blob/master/src/lsm303dlh/lsm303dlh.cxx#L119
2. https://github.com/intel-iot-devkit/upm/blob/master/src/lsm303dlh/lsm303dlh.cxx#L121

bjbeare:
Please post the patch so that the maintainer can accept/reject/comment.

jontrulson:
Hi,

These were early drivers, where the authors (myself included) ASSUMED that since each I2C transaction (at the hardware bus level) starts with the device address, then the mraa_i2c_address() call must be called before every transaction.  This was due to a misunderstanding of exactly what the mraa_i2c_address() call actually did under the hood.

As you notice, this is not required except in very specific circumstances.  The mraa_i2c_address() function (or C++ .address() method) caches the address specified in the I2c context, which is then used when issuing "real" I2C transactions.  Calling it repeatedly as you notice is not necessary, but also does not affect the I2C transaction on the hardware level either.  As you point out, newer drivers do not do this.

[...]

PR #554, opened 2017-05-17
smartdrive: fixed x and y

neuberfran:
Signed-off-by: neuberfran <neuberfran@bol.com.br>


neuberfran:
Please, tell me why files smartdrive.cxx and smartdrive.hpp in upm sensor not have new changes yet

Propanu:
Does this code work correctly for your setup? The only significant changes I see are [this](https://github.com/intel-iot-devkit/upm/pull/554/files#diff-270a402cdeae1f945260be6088d61817R187) and [this](https://github.com/intel-iot-devkit/upm/pull/554/files#diff-270a402cdeae1f945260be6088d61817R191), pointing to your motors being swapped compared to the image on the first page of the [user manual](http://www.mindsensors.com/index.php?controller=attachment&id_attachment=121). This should be fixed in your application code and not in the driver.

neuberfran:
You are wrong. I tested atual code in two intel edison boards with yocto/node.js scripts (my) and not worked correctly. With my code (smartdrive.cxx and smartdrive.hpp) My scripts worked. Please change.

Propanu:
Can you paste the node.js script you are using with the library here?

There are several issues with your updated code that did not pass our review, namely:
- Commit message needs to be updated, I used "x" and "y" as placeholders for actual description of changes.
- Author should not be removed and replaced, it's fine if you want to add yourself to the list though.
- There are a couple of instances where you replace a symbol with the exact same numerical value in the .hpp file, e.g. `CMD_S == 0x53`. This is not needed.
- In some places you use a hard-coded '5' when calling `writeArray()`, but the array is larger than 5 bytes, throwing away the rest of the command.
- You shouldn't make repeated calls to `i2c.address` since it's saved in the context (see #547).
- Formatting/indentation should match existing code (you have extra spaces, tabs, etc.)

neuberfran:
var smart = require('jsupm_smartdrive');
var smarttest = smart.SmartDrive;
var smarttest = new smart.SmartDrive(0);
var teste = smarttest.GetBattVoltage();

setTimeout(function(){
    smarttest.Run_Seconds(0x02,0x00, 90, 70, true, 0x01 );

    console.log(teste);
    setInterval(function(){
           smarttest.StopMotor(0x02,0x02);
    }, 4000);
}, 4000);

**and**

ar smart = require('jsupm_smartdrive');
var smarttest = new smart.SmartDrive(0);
var teste = smarttest.GetBattVoltage();

setTimeout(function(){
    smarttest.Run_Seconds(0x03,0x00, 90, 70, true, 0x01 );

    console.log(teste);
    setInterval(function(){
           smarttest.StopMotor(0x03,0x02);
    }, 4000);
}, 4000);

**I do not see problems in leaving the author's name**
Oussema Harbi <oussema.elharbi@gmail.com>

Propanu:
With your code changes and 1st test script, which motor turns M1 or M2? Refer to the picture on the first page [here](http://www.mindsensors.com/index.php?controller=attachment&id_attachment=121). It may be possible that the firmware of the motor controller has them backwards.

I'm assuming with the 2nd test script both of them turn as expected?

neuberfran:
With the first script run motor 2 (0x02). But if  run a second script Immediately after the first, run  only motor 2 and stop after 2 seconds. After that, yes, the two motors. I change de smarttest do smarttestb in first script but not solved. I stay usaing my smartdrive.cxx and my smartdrive.hpp codes in this casa. The smartdrive.cxx and smartdrive.hpp of the upm sensor I try i last month and not worked

neuberfran:
Today, 22:00 PM I stay trying use atual smartdrive atual sensor, and I have issues:

First I do this: **cmake ..// make smartdrive // make install** 

tela2doc

After I try run scripts above and code of the /usr/examples/c++ with ID_1  because this: (https://drive.google.com/file/d/0B9gPdWq4OH-XaUpfekFyLXdXeDQ/view?usp=sharing)

**_### But motors never run_** with _BOTH option the light red in smartdrive physic ON
tela4c

neuberfran:
The code that is currently in github is definitely wrong. Look at this:

image

We need to correct

Propanu:
@neuberfran, the [python library provided by openelectrons](https://pypi.python.org/pypi/SmartDrive/1.0) matches the current code in UPM:
```py
            if ( (motor_number & 0x01) != 0 ):
                array = [speed, duration, 0, ctrl]
                self.writeArray( self.SmartDrive_SPEED_M1, array) 
            if ( (motor_number & 0x02) != 0 ) :
                array = [speed, duration, 0, ctrl]
                self.writeArray( self.SmartDrive_SPEED_M2, array)
```
What may be happening in your case, is that you build the new library after it was patched, and install it under `/usr/local/lib` on your Edison. However, the old library is still under `/usr/lib` and that's the one your example links against. This is further hinted to from the error messages thrown by dlopen. Can you check the version on that and confirm it's 1.0 or newer?

Alternatively, run
```
opkg update
opkg upgrade mraa upm
```
On your Edison and it should grab the latest for you. Then try your test script again.

neuberfran:
https://github.com/intel-iot-devkit/upm/pull/557

neuberfran:
Driver funcionou no Android Things DP3/intel edison/jcenter/gradle

PR #585, opened 2017-08-01
Added first cluster of JSON sensor files(40 files).

Msgtfrank:
Added this first cluster of 40 JSON files

### DIFF
--- a/src/apds9002/apds9002.json
+++ b/src/apds9002/apds9002.json
@@ -1,0 +1,61 @@
+{
+    "Library": "apds9002",
+    "Description": "Seeed luminance sensor library",
+    "Sensor Class": {
+        "APDS9002": {
+            "Name": "Seeed luminance sensor",
+            "Description": "This is the UPM Module for the apds9002, Seeed luminance sensor. This sensor offers a luminance sensor solution that has a responsitivity close to that of the human eye. Operates at 3.3 and 5 volts on a Seeed standard breakout board.",
+            "Aliases": ["apds9002", "Grove - Luminance Sensor"],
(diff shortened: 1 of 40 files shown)


src/apds9002/apds9002.json
```diff
@@ -0,0 +1,61 @@
...
+                    "unit": "ms",
+                    "typ" : 0.8,
+                    "max" : 2
+                }
+            },
+            "Platforms": {
+                "Intel Edison": {
+                    "Notes": ["Might require Seeed breakout board"]
+                },
+                "Arduinon 101": {
```
pylbert: Spelling - Arduino x 43

Msgtfrank: Oh wow I didn't notice that till now. It's even in the templete we've been using, which is based on another sensor. I'm a bit new to being a contributor in this, should I change the files and submit a new pull request or?...

pylbert: One way is to fix the spelling, do a commit --amend, and then force push.

Msgtfrank: Aha! Alright. I'll get that done. do I need to do another pull request to do that, or will the push apply to this request?

dnoliver:
I think we should put some level of validation to those packages. We could use [jsonlint](https://github.com/zaach/jsonlint), and if we have well defined schema (i.e. every json must have the "Library" ) we can code an automated test to validate that for each of them.

src/apds9002/apds9002.json
Propanu: When you update your branch it will automatically update the PR too.

PR #587, opened 2017-08-07
zfm20: Use mraa::Uart instead of tty

pylbert:
Updated the ZFM20 class to use UART functionality provided through the
mraa::Uart class instead of using the UART directly.

    * Switch to mraa::Uart
    * Updated examples
    * Added a common.i to minimize duplication
    * Removed pointers from C++ functions where references are
      preferable
    * Removed dependency on termios
    * Added some typedefs to handle pass-by-reference


src/zfm20/zfm20.cxx
```diff
@@ -33,154 +33,53 @@ using namespace std;
...
 
-  tcdrain(m_ttyFd);
+  m_uart.flush();
 
   return rv;
 }
 
-bool ZFM20::setupTty(speed_t baud)
+bool ZFM20::setupTty(uint32_t baud)
 {
```
jontrulson: If you are amenable to API changes, I would just remove this function and have the baudrate, passed as a standard int, to the ctor, and then set it there.  Or, this function could just be renamed to setBaudrate() instead (but still accepting a standard int rather than int32_t).  As you know by now, almost all of what setupTty() did is handled by MRAA now anyway.

src/zfm20/zfm20.cxx
```diff
@@ -231,12 +130,12 @@ uint32_t ZFM20::getMillis()
   gettimeofday(&now, NULL);
```
jontrulson: I'll bet this timing code could be replaced by a upm_clock_t and it's functions.  Not sure if you want to do that now though.

src/zfm20/zfm20.cxx
```diff
@@ -540,10 +439,10 @@ uint8_t ZFM20::deleteDB()
   return rPkt[9];
 }
 
-uint8_t ZFM20::search(int slot, uint16_t *id, uint16_t *score)
+uint8_t ZFM20::search(int slot, uint16_t &id, uint16_t &score)
 {
```
jontrulson: Have you verified the device with this?  I always avoided exposing references to SWIG, due to [this SWIG documentation](http://www.swig.org/Doc1.3/SWIGPlus.html#SWIGPlus_nn18)

It says references are just converted to pointers and there are apparently issues with const'ness (due to the fact that in SWIG they are converted to pointers).  But if the device still works ok in all languages, go for it.

But, if it works...

src/zfm20/zfm20.cxx
pylbert: I will remove the call altogether.   Seems to make a bit more sense rather than just wrapping the mraa call with a different return value.

src/zfm20/zfm20.cxx
pylbert: Will do.

src/zfm20/zfm20.cxx
pylbert: Verified with python using typemaps to handle the references.  IMO, this is easier (at least in python) since it returns a list (example search(1) = [0, 1, 80]) but I will double-check for java and javascript.  The current pointer implementation in C++ is weird and missing any check before de-referencing the pointers.

jontrulson: >  IMO, this is easier (at least in python) since it returns a list (example search(1) = [0, 1, 80])...

I'm confused... search accepts 3 args and returns a uint8_t... So I'm not sure how you call search(1), and how you get a tuple back...?

pylbert: I use typemaps to handle the references.  The wrapped method takes 1 argument and returns a list (not a tuple).  Details below.

```
/* Make it easy to use the methods which take a uint& id and score */
%include "typemaps.i"
%apply uint16_t &OUTPUT {uint16_t &id};
%apply uint16_t &OUTPUT {uint16_t &score};
```

Generates CXX with uint16_t temp variables for the two references (that I've defined at OUTPUT). 
 These temp variables get appended to the pyObject return:

```
SWIGINTERN PyObject *_wrap_ZFM20_search(PyObject *SWIGUNUSEDPARM(self), PyObject *args) {
...
uint16_t temp3 ;
uint16_t temp4 ;
...
arg3 = &temp3;
arg4 = &temp4;\
...
result = (uint8_t)(arg1)->search(arg2,*arg3,*arg4);
...
if (SWIG_IsTmpObj(res3)) {
  resultobj = SWIG_Python_AppendOutput(resultobj, SWIG_From_unsigned_SS_short((*arg3)));
...
if (SWIG_IsTmpObj(res4)) {
  resultobj = SWIG_Python_AppendOutput(resultobj, SWIG_From_unsigned_SS_short((*arg4)));
```

Which is wrapped like so in pyupm_zfm20.py:

```
def search(self, slot):
    return _pyupm_zfm20.ZFM20_search(self, slot)
```

Usage:

```
>>> import pyupm_zfm20
>>> x = pyupm_zfm20.ZFM20(0)
>>> x.setupTty(57600)
True
>>> x.generateImage()
0
>>> x.image2Tz(1)
0
>>> x.generateImage()
0
>>> x.image2Tz(2)
0
>>> x.search(1)
[0, 1, 78]

```

pylbert: The drawback is the doxygen-generated docs won't reflect this :(

jontrulson: I see - hadn't paid attention to the interface files :)  

In addition to documentation, there's also the issue that the API is significantly different between C++ and SWIG languages...

pylbert:
Closed in 680649ba6f822e8a7f55097b18b86bf4b60b29d4

PR #589, opened 2017-08-10
Sensor jso ns2

Msgtfrank:


malikabhi05:
Merged, thanks a lot for the quick response!!

Issue #593, opened 2017-08-17
pyupm_bno055 module usage issue


karthikreddyk:
Hi, 

I am using above captioned module to capture IMU data from BNO055 , a 9 axis IMU from Bosch/Adafruit.
From time to time I get this message: bno055_read_reg: mraa_i2c_read_byte_data() failed, returning 0
and all of a sudden my accelerometer and angle outputs shoot up.

For example from say lets 1 m/s^2 to the X and Y component accel values displayed below:

accel_Data:
(-195.33999633789062, 137.60000610351562).

I am presuming the issue is because of lack of proper exception handling. Any help is greatly appreciated.
Sincerely,
Karthik Kambam

jontrulson:
Can you tell me what function was calling bno055_read_reg()?  Typically after you setup/calibrate the chip, you will then call update() in a loop, and then use the various getters, like getLinearAcceleration() to obtain the data of interest.

The update() routine is the only place this can happen during _normal_ operation, and the only thing it's used for in update() is obtaining the chip temperature.  So, during an update() call, the temperature would return 0.0 in this case.  It would not affect acceleration, gyro, or anything else (different low level routines are used for those).  So, I suspect it's just a symptom of another more serious problem.

We could fix this particular problem by modifying the code to check the result of every 'single register' read, and throw an exception, but the end result should still be the same in your case - since that routine is not used to obtain fusion and non-fusion data (accelerometer, etc)...

jontrulson:
I guess one other obvious question - are you using the chip temperature (via getTemperature()) for anything important in your code?

karthikreddyk:
Hi Jon,

Thank you for your prompt response. 
I am not calling getTemperature() at all.
In a loop I call update() first , then I call getLinearAcceleration(), and then getEulerAngles().
sometimes I get the message in sysout:
bno055_read_reg: mraa_i2c_read_byte_data() failed, returning 0.
In such cases, the acceleration data from getLinearAcceleration would be an impossible value like the one printed below:
accel_Data:
(-195.33999633789062, 137.60000610351562) - Z component not shown.
I am getting an impossible value from getEulerAngles() too.

The message: "bno055_read_reg: mraa_i2c_read_byte_data() failed, returning 0" makes me believe that the following module gets called when I call getLinearAcceleration() and getEulerAngles():

uint8_t bno055_read_reg(const bno055_context dev, uint8_t reg)
{
    assert(dev != NULL);
[...]

PR #595, opened 2017-08-19
bno055: enhance error detection and propagation

jontrulson:
This patch reworks error handling in the C driver to more reliably detect
errors, and for C++, throw exceptions when they are detected.

The C++ API is unchanged aside from the fact that more methods will
throw an exception on errors now.

This addresses the error handling deficiencies reported in Issue #593.


malikabhi05:
Thanks a lot for this @jontrulson , merged!

PR #609, opened 2017-10-04
testing cordova bindings generation

dnoliver:
This pr is should not be merged, sis just to share the implementation code of cordova-plugin generation


dnoliver:
@vintummala @Propanu 

I just copied and pasted the files shared in google drive, and execute them. The upm-cordova-converted/main.js file requires nodejs to be executed, and you need java to build the Java upm modules. 

I do not know if you just need the Java modules (the "standard" ones), or the Java for Android Things modules. If somebody can add the developer that implemented those files to this PR, please do so I have more information about this.

Propanu:
@dnoliver I think you need the `io.mraa.at` and `io.mraa.at.upm` modules.

dnoliver:
@Propanu now it will build cordova plugins using upm modules built for android things. 

I think this is done from my side, I will just need to get this plugins tested (@vintummala told me that there is a team that could help with that task). 

Then we could just refactor the code and add build automation for it

PR #613, opened 2017-10-13
CPack: Include latest CPackDeb for building IPKs

pylbert:
Replaced CPackDeb.cmake with the latest version which supports naming
the debian packages with a .ipk suffix.  Included CMake's 3-Clause
License for redistribution.


dnoliver:
ipk generation is throwing a segmentation fault. Should this change fix that?

pylbert:
The CPack binary segfaults on various string methods (usually in cmCPackDebGenerator.cxx where GetOption returns a NULL to a string constructor).  This version of CPackDeb creates an IPK and does not cause the segfault.

Issue #617, opened 2017-11-12
[Question] Support for LIS3DH and MAX7219/21


alext-mkrs:
@propanu, @w4ilun, I've got my hands on those two subj modules and noticed there seems to be no support for either of them (there's LIS2DS12, which is very close, but still a bit different from LIS3DH and the latter seems to be more widespread being cheaper) in UPM.

Is that correct? If so, are you interested in getting those in, I could give it a shot if no one works on that or plans to.

w4ilun:
@alext-mkrs I don't think those 2 modules are in development, would be great if you could add them!

alext-mkrs:
Good, thanks for the information :) I'll follow the standard PR flow then, when it's ready.

PR #622, opened 2017-12-30
lis2ds12.{h,hpp}: add missing parameter description

alext-mkrs:
I'm working on a LIS3DH sensor module as discussed earlier in #617 and use @jontrulson's [nice] lis2ds12 one as the base. Noticed a missing parameter description, so here goes the fix.


Propanu:
Merged, thanks @alext-mkrs!

PR #623, opened 2018-01-06
lis3dh: add sensor support based on lis2ds12 module

alext-mkrs:
Adding STMicro LIS3DH sensor support. This module is based on the one for LIS2DS12 (thanks, @jontrulson !), but as sensors are noticeably different, the contents underwent major rework. I've also run `clang-format` on all C/CXX/H/HPP module files and C/C++ examples.

Examples and basic API are left the same.

Tested on Intel Edison with Arduino board using both I2C and SPI.

One particular difference to note is that I removed the `accFactor` coefficient, as it is effectively cancelled out right away in `lis2ds12_get_accelerometer()` calculations: https://github.com/intel-iot-devkit/upm/blob/master/src/lis2ds12/lis2ds12.c#L455-L458, so seems to be of no use and I couldn't find any reason for preserving it. Line 455 adds it to the numerator, line 458 - to the denominator, cancelling it out. In addition, the logic of the necessary conversion itself doesn't call for such an additional coefficient IMHO, which is additionally confirmed by the calculation example in [LIS2DS12 appnote](http://www.st.com/resource/en/application_note/dm00218428.pdf), see section 3.5.1 "Example of output data", page 16.

If you think those are actually necessary, I'd be grateful for a clarification. If these are indeed surplus - I'd be glad to submit a separate PR for removing them from `lis2ds12` module as well.

In addition, as we anyway return values in "g" and not "milli g", I also removed the divison by 1000 in `lis3dh_get_accelerometer()` and simply set proper scaling factor in "g" right away in `lis3dh_set_full_scale()`.

### DIFF
--- a/examples/c++/lis3dh.cxx
+++ b/examples/c++/lis3dh.cxx
@@ -1,0 +1,81 @@
+/*
+ * Author: Alex Tereschenko <alext.mkrs@gmail.com>
+ * Copyright (c) 2018 Alex Tereschenko.
+ *
+ * Based on LIS2DS12 example by
+ * Author: Jon Trulson <jtrulson@ics.com>
+ * Copyright (c) 2017 Intel Corporation.
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining
+ * a copy of this software and associated documentation files (the
+ * "Software"), to deal in the Software without restriction, including
+ * without limitation the rights to use, copy, modify, merge, publish,
+ * distribute, sublicense, and/or sell copies of the Software, and to
+ * permit persons to whom the Software is furnished to do so, subject to
+ * the following conditions:
+ *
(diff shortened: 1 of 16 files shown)


jtrulson:
> If you think those are actually necessary, I'd be grateful for a clarification. If these are indeed surplus - I'd be glad to submit a separate PR for removing them from lis2ds12 module as well.

Nope, I think you are correct there.  Of course, please make sure the changes work -- they should, but I've been unpleasantly surprised before. :)

alext-mkrs:
Thanks! Yeah, they do work, I've surely tested that :)

alext-mkrs:
So, any comments or do you think we can merge this, @propany, @pylbert?

pylbert:
Great add!  Merged.

alext-mkrs:
Thanks!

PR #633, opened 2018-03-14
Cordova

pylbert:


cmake/modules/FindCordova.cmake
```diff
@@ -0,0 +1,8 @@
+find_program(UPM_CORDOVA_BINDING NAMES upm-cordova-binding
```
dnoliver: Cordova is an actual npm package (it was the foundation of the Intel XDK), I think it is more appropriate to name this `FindUpmCordovaBindings.cmake`

cmake/modules/FindCordova.cmake
```diff
@@ -0,0 +1,8 @@
...
+)
+
+if(UPM_CORDOVA_BINDING)
+    message(STATUS "The npm package to build cordova bindings found at ${UPM_CORDOVA_BINDING}")
```
dnoliver: Maybe we should add a version declaration here as well, as Yuidoc do https://github.com/intel-iot-devkit/upm/blob/master/cmake/modules/FindYuidoc.cmake#L16 and check for the version as well

docker-compose.yaml
```diff
@@ -113,6 +114,13 @@ services:
       - BUILDTESTS=OFF
     command: bash -c "./scripts/build-android.sh"
 
+  cordova:
+    extends: all
+    environment:
+      - BUILDSWIGJAVA=ON
+      - BUILDCORDOVA=ON
+    command: bash -c "npm install -g upm-cordova-binding && ./scripts/run-cmake.sh && cd build && make -j8 && make -j8"
```
dnoliver: We should add this to the Docker Image! instead of running an installation each time you generate the bindings, the upm-cordova-bindings will be installed already in the upm-all image that CI downloads. 

Also with this you depend on a working internet connection to run the cordova bindings generation (which is not always possible for developers in other countries  )

pylbert: Done.

cmake/modules/FindCordova.cmake
pylbert: Done.

pylbert:
We're still awaiting a fix for the cordova binding generator.

dnoliver:
should we put that repository with the code of the cordova binding generator into this organization?

pylbert:
Closing, no longer needed.

PR #635, opened 2018-03-18
Fix kx122 python

AntAgna:
Many of the methods were unusable in Python through SWIG due to these problems:

- uint was seen as an 'object' by SWIG
- using pointers to return values

### DIFF
--- a/src/kx122/kx122.hpp
+++ b/src/kx122/kx122.hpp
@@ -82,10 +82,10 @@ namespace upm{
       /**
       Gets who am i value of the sensor.
 
-      @param data Pointer to a uint8_t variable to store the value.
+      @return Who am I value of the sensor.
       @throws std::runtime_error on failure.
       */
-      void getWhoAmI(uint8_t *data);
+      uint8_t getWhoAmI();
 
       /**
       Gets raw accelerometer data from the sensor.
@@ -339,10 +339,10 @@ namespace upm{
 
       See datasheet for more details.
 
-      @param data Pointer to a uint8_t variable to store the value.
+      @return Interrupt source value.
       @throws std::runtime_error on failure.
       */
-      void getInterruptSource(uint8_t *data);
+      uint8_t getInterruptSource();
 
       /**
       Clears latching interrupts (Wakeup, Data Ready).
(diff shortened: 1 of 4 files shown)


pylbert:
@AntAgna, good changes all-around.  I went ahead and merged the TSL2561 fix in a separate commit (83f541a5db8747b1d2a7c019655071b2f4c7aa9e) since both your PR's include this and a few additional items needed to be updated.

I also think we may be able to handle the uints for all python modules with an apply - can you see if this would work for you (d40019905551a8857f64bec030da793bec845cf)?

The additional pointer types are definitely problematic and I would generally not want to see these in C++.  I like your stl vector change.  Can you update your PR to include vector returns for all the C++ methods which take/return pointers?

src/kx122/kx122.hpp
```diff
@@ -486,20 +486,16 @@ namespace upm{
       @param z_array Pointer to an floating point array to store the z-axis data. Can be set to NULL if not wanted.
       @throws std::runtime_error on failure.
       */
-      void getRawBufferSamples(uint len, float *x_array, float *y_array, float *z_array);
-
+      void getRawBufferSamples(int len, float *x_array, float *y_array, float *z_array);
+
```
Propanu: Please remove these spaces.

AntAgna:
I have added (d400199) instead of replacint uint. I don't have my test platform available right now but I think it should work.

src/kx122/kx122.hpp
```diff
@@ -511,6 +503,9 @@ namespace upm{
       //Device context
       kx122_context m_kx122;
 
+      //Maximum number of samples that can be stored in the buffer of the KX122
+      const static uint MAX_SAMPLES_IN_BUFFER = 681;
```
pylbert: SWIG seems to have trouble with the static member variable.  Since this is private, does it need to be static?

AntAgna: If it is not static, it will be added inside the object.
Would a enum value be better ?

Propanu: I think a `#define` will work just fine, we use them throughout and SWIG seems ok with those.

AntAgna: Changed it to use a #define

pylbert:
Thanks @AntAgna!

Closing - merged in b55501e327837bfaa20c745c1b0a06bd0234a624, e99f1d73fda6c893fb2fe2f950d6d9888a359dba, and fc56e5604831723da4f255ca3d6ea5f3f33d2d5b.  I updated the apichanges a bit as well.

PR #645, opened 2018-05-08
Replace strncpy with memcpy

kraj:
gcc8 detects that strncpy is overwiritng the null terminating character
the source strings are already initialized to 0 so memcpy would do the same
job

Fixes
rn2903.c:153:5: error: 'strncpy' output may be truncated copying 16 bytes from a string of length 511 [-Werror=stringop-truncation]
     strncpy(dev->hardware_eui, dev->resp_data, RN2903_MAX_HEX_EUI64);
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/ecezo/ecezo.c
+++ b/src/ecezo/ecezo.c
@@ -488,7 +488,7 @@ int ecezo_send_command(const ecezo_context dev, char *cmd, char *buffer,
     // our write buffer
     char writeBuffer[ECEZO_MAX_BUFFER_LEN];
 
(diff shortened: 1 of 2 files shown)


Propanu:
Merged with 8a617aa9044cbe5731b32d79437bd6a9ffcf2602, thanks!

PR #650, opened 2018-06-07
Logger

stefan-andritoiu:


include/upm_logger.hpp
```diff
@@ -0,0 +1,131 @@
...
+  protected:
+    std::ostringstream os;
+  private:
+    UPM_LOGGER(const UPM_LOGGER&);
+    UPM_LOGGER& operator=(const UPM_LOGGER&);
+  private:
+    // Track logLevel for this particular instance of UPM_LOGGER
+    UpmLogLevel logLevel;
+    std::ofstream logStream;
+    static std::mutex logMutex;
```
pylbert: The implementation of this variable below will cause problems if included from two different translation units.  One way to overcome this would be:

```
static std::mutex& logMutex()
{
    static std::mutex _logMutex;
    return _logMutex;
}
```

include/upm_logger.hpp
```diff
@@ -0,0 +1,131 @@
...
+  public:
+    static const std::string& getLogLevelName(UpmLogLevel level);
+  private:
+    void write(std::ostringstream& os);
+  };
+
+  std::mutex UPM_LOGGER::logMutex;
```
pylbert: The implementations from here on down will most likely also cause problems.  I would include them in the class definition.

include/upm_logger.hpp
```diff
@@ -0,0 +1,131 @@
...
+
+#include <sstream>
+#include <fstream>
+#include <string>
+#include <chrono>
+#include <thread>
+#include <mutex>
+
+/* Helper macro for logger utility class. */
+#define UPM_LOG(log_level) \
```
pylbert: Maybe we extend these further and provided helpers for the different loglevels as well?

pylbert:
I like this implementation - UPM needs a unified logging scheme.

- Is there a means to use this in the UPM C source?
- Let's add a set of unit tests to verify everything works as expected and ensure coverage.
- Make sure to document the logger class/methods.
- It would be nice to have some control over the target file (whether that's /dev/stdout, ./localfile.log, /tmp/someotherfile.log)

Some formatting ideas below (it would be nice to have an optional classname in there if a method belongs to a class).  If we're talking one log file for all binaries built from UPM, then we most likely would need to add either a pid or executable name to differentiate them.

```
[2018.06.07 10:54:40.638] [WARNING] filename.cxx:10 (ClassName::)function "My error message to log"
[2018.06.07 10:54:40.638] [DEBUG  ] filename.cxx:10 (ClassName::)function "My debug message to log"
```

If this logging implementation existed in mraa, both projects could benefit.  Thoughts?

PR #659, opened 2018-07-27
BH1749: C/C++ driver implementation with examples

Issam-b:
This pull request contains the implementation for the ROHM BH1749 color sensor driver. Samples examples for all 5 supported languages are included.


Issam-b:
Alright, that's good. About those minor changes, it's OK to correct them when merging, but can you point them out, since I'll be adding another sensor soon, and it's better not to have the same issues again!

Propanu:
For the JSON file it's mostly a matter of formatting, some of the entries there can be kept on a single line for readability, looks good otherwise.

On the other hand, the .hpp file needs a "library block" so that the page generated from it gets linked in the documentation (see [this page](https://iotdk.intel.com/docs/master/upm/modules.html)). I recommend using an existing sensor driver as a guide, for instance [TCS3414](https://github.com/intel-iot-devkit/upm/blob/master/src/tcs3414cs/tcs3414cs.hpp) would be a good one. You can find more details on this in the document [here](https://github.com/intel-iot-devkit/upm/blob/master/docs/documentation.md#doxygen-tags). If you want to check how it looks before submitting a PR, it helps to enable documentation during builds. This does have some additional dependencies (explained in building.md file) thus it's understandable if you decide to skip this.

Propanu:
Merged with some minor formatting changes. Thanks!

PR #666, opened 2018-09-09
Use IdleNotificationDeadline as deprecated IdleNotification has been …

nullr0ute:
…removed

In v8 API 6.5 the long deprecated IdleNotification has been removed so we
need to use IdleNotificationDeadline as nodejs 10.x uses newer v8.

Signed-off-by: Peter Robinson <pbrobinson@gmail.com>


nullr0ute:
Fixes issue 665

PR #679, opened 2019-03-13
Omron 2jcie bu01 usb

omron-ankit:
Initial implementation of Omron's USB type Environment Sensor using USB to UART interface


Propanu:
Merged, thank you! Please consider consolidating this further by implementing interfaces, adding additional examples and perhaps refactoring under a single library.

PR #680, opened 2019-03-13
Omron 2jcie bu01 ble

omron-ankit:
Initial implementation of Omron's Environment sensor using BLE interface.


Propanu:
Merged with some additional changes, thank you! Please consider consolidating this further by implementing interfaces, adding additional examples and perhaps refactoring under a single library. Also, this is not currently available in Java.

PR #710, opened 2025-11-23
CMakeLists: update min version to 3.10 for CMake >= 4.0 support

Ansuel:
Update minimum version of CMake to 3.22 for CMake >= 4.0 version
support.

New CMake require 3.5 as the minimum version with that increased to 3.10
in the next CMake release.

Drop old redundant CMake condition not needed anymore due to new version
requirement.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 2.8.11)
+cmake_minimum_required (VERSION 3.22)
 project (upm)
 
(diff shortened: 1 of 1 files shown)


alext-mkrs:
Thanks for your contribution. I think it would be better to follow the PR I had for mraa and use the same version here for uniformity, check it out here: https://github.com/eclipse/mraa/pull/1148.

Ansuel:
@alext-mkrs thanks for the check, now should be better

